PR #168, opened 2019-04-14
3d tiles: binary parsers for tiles

ibgreen:
Binary parsers for the main 4 tile formats, passing basic unit tests using Cesium test tiles.


ibgreen:
@xintongxia @lilleyse FYI

PR #200, opened 2019-05-04
Separate workers from normal parser

xintongxia:


modules/arrow/src/index.js
```diff
@@ -1,4 +1,21 @@
...
 
 export {default as ArrowTableBatch} from './lib/arrow-table-batch';
+
+// DEPRECATED
+
+// eslint-disable-next-line import/first
+import ArrowLoader from './lib/parse-arrow-sync';
+
+const DEPRECATION_WARNING = `\
+ArrowWorkerLoader must be imported from @loaders.gl/arrow/arrow-loader. \
```
Pessimistress: `worker-loader`

modules/las/worker-loader.js
```diff
@@ -1,5 +1,5 @@
 // The bundled worker is imported as an inline string
-import worker from '../dist/las-loader.worker.js';
+import worker from './dist/las-loader.worker.js';
```
Pessimistress: This file will not be transpiled and break any environment that does not support esm style `import`.

Pessimistress:
Browser test fails.

PR #266, opened 2019-06-20
Remove argument passed to TextEncoder that causes console.warn() to be called

twojtasz:
The culprit for the `console.warn()`  is here:
 https://github.com/uber-web/loaders.gl/blob/master/modules/polyfills/src/text-encoding/encoding.js#L1125


twojtasz:
will do

PR #282, opened 2019-06-28
Use wgs84 transform for royal exhibition

loshjawrence:
This overrides the transform for royal exhibition until the actual tileset.json transform is fixed, after which, the override can be removed. 

Also includes some fixes to bounding-volume to use real oriented-bound-box which is part of the way to using real distance checks.


modules/3d-tiles/src/tileset/helpers/bounding-volume.js
```diff
@@ -72,21 +65,30 @@ export function createBoundingVolume(boundingVolumeHeader, transform, result) {
...
-
-  const center = new Vector3(box[0], box[1], box[2], scratchCenter);
-  const halfAxes = new Matrix3(); // Matrix3.fromArray(box, 3, scratchHalfAxes);
-
-  // Find the transformed center and halfAxes
-  center = Matrix4.multiplyByPoint(transform, center, center);
-  const rotationScale = Matrix4.getRotation(transform, scratchMatrix);
-  halfAxes = Matrix3.multiply(rotationScale, halfAxes, halfAxes);
+  const center = new Vector3(box[0], box[1], box[2]);
+  let halfAxes = new Matrix3(box.slice(3, box.length));
```
ibgreen: You never use this value

examples/3d-tiles/package.json
```diff
@@ -13,10 +13,10 @@
     "@loaders.gl/3d-tiles": "^1.0.3",
     "@loaders.gl/core": "^1.0.3",
     "@luma.gl/addons": "^7.1.0",
-    "math.gl": "^3.0.0-alpha.3",
-    "@math.gl/geospatial": "^3.0.0-alpha.3",
-    "@math.gl/culling": "^3.0.0-alpha.3",
+    "@math.gl/culling": "^3.0.0-alpha.5",
```
ibgreen: Nit: I think xintong published an alpha.6.

examples/3d-tiles/tile-3d-layer.js
```diff
@@ -50,6 +50,29 @@ export default class Tile3DLayer extends CompositeLayer {
     let tileset3d = null;
     if (tilesetUrl) {
       const tilesetJson = await load(tilesetUrl);
+
+      if (tilesetUrl.includes('RoyalExhibitionBuilding')) {
```
ibgreen: You should be able to cut this now.

modules/3d-tiles/src/tileset/helpers/bounding-volume.js
loshjawrence: Not sure what you mean

ibgreen:
@loshjawrence I can land this as soon as you have rebased to master

modules/3d-tiles/src/tileset/helpers/bounding-volume.js
ibgreen: You are assigning `halfAxes` on line 69, and overwriting it on line 73, as far as I can tell, you don't use the value in between.

loshjawrence:
@ibgreen 
Having some trouble compliing after the merge something about babel/polyfill:
image

ibgreen:
>  about babel/polyfill:

after rebasing, it is a good idea to run yarn in the example (and ideally also yarn bootstrap in root).

loshjawrence:
the bootstrap did it, thanks.

modules/3d-tiles/src/tileset/helpers/bounding-volume.js
loshjawrence: there's a multiplyRight on 82 that uses it

loshjawrence: There's a multiplyRight on 82 that uses it.

loshjawrence:
@ibgreen I think this is good to go, unless the halfAxes multiplyRight usage on 82 could be made more readable or something.

modules/3d-tiles/src/tileset/helpers/bounding-volume.js
ibgreen: Nevermind... I looks to me like it uses the value you assign on line 72. The value you initialize with on line 68 is overwritten and never used.

PR #283, opened 2019-07-01
Real distance sse traversal

loshjawrence:
Using actual* distances instead of zoom for sse traversal. Added a onTilesetLoad callback so that the camera moves to the tileset when its json is loaded instead of when the root tile with content. If the camera isn't moved to the tileset traversal will cull the root (and so on).

There's an isRoyal boolean, needed until https://github.com/uber-common/deck.gl-data/pull/11 is merged.

*Almost, the long lats are pulled from vieport but the the height that's used is a zoom re-mapping since I didn't see anything that looked like a meter height relative to some ellipsoid surface.


examples/3d-tiles/tile-3d-layer.js
```diff
@@ -52,9 +54,9 @@ export default class Tile3DLayer extends CompositeLayer {
       tileset3d = new Tileset3D(tilesetJson, tilesetUrl, options);
 
       // TODO: Remove these after sse traversal is working since this is just to prevent full load of tileset and loading of root
-      // The alwaysLoadRoot is better solved by moving the camera to the newly selected asset.
       tileset3d.depthLimit = this.props.depthLimit;
-      tileset3d.alwaysLoadRoot = true;
+      // TODO: remove after data PR merge
```
ibgreen: Merged!

loshjawrence:
Looks like its pulling in the new tileset.json. Going to remove the isRoyal boolean.

modules/3d-tiles/src/tileset/helpers/bounding-volume.js
```diff
@@ -43,22 +42,14 @@ export function createBoundingVolume(boundingVolumeHeader, transform, result) {
...
       (minHeight + maxHeight) / 2
     );
+    result = result || {};
     Object.assign(result, boundingVolumeHeader, {center});
 
     return result;
     // return createRegion(boundingVolumeHeader.region, transform, this._initialTransform, result);
   }
   if (boundingVolumeHeader.sphere) {
-    // The first three elements define the x, y, and z values for the center of the sphere in a right-handed 3-axis (x, y, z)
```
ibgreen: As mentioned, I do think the code to extract an initial suggested "view state" should be somewhere in the tileset rather than in the app (there are three type of bounding boxes, the app should not need to deal with that...) . 

Perhaps add a separate method `_getCartographicCenterAndZoom`?

examples/3d-tiles/tile-3d-layer.js
```diff
@@ -87,17 +89,21 @@ export default class Tile3DLayer extends CompositeLayer {
...
-    const maxZoom = 21;
+    const minZoom = 16;
+    const maxZoom = 20;
     const zoomMagic = 10000; // royalexhibition
     let zoomMap = Math.max(Math.min(zoom, maxZoom), minZoom);
     zoomMap = (zoomMap - minZoom) / (maxZoom - minZoom);
     let expMap = 1 - Math.exp(-zoomMap * 6); // Use exposure tone mapping to smooth out the sensitivity in the zoom mapping
     expMap = Math.max(Math.min(1.0 - expMap, 1), 0);
     const distanceMagic = expMap * zoomMagic;
+    const heightMagic = expMap * zoomMagic + 16; // I think royal was tiled at a height of 16
```
xintongxia: Is this `heightMagic` only for royal case?

examples/3d-tiles/app.js
```diff
@@ -206,30 +215,31 @@ export default class App extends PureComponent {
...
-        console.warn('center was not pre-calculated for the root tile');
-      } else {
-        scratchLongLat.copy(center);
-        if (isRoyal || name === 'TilesetPoints') {
-          Ellipsoid.WGS84.cartesianToCartographic(center, scratchLongLat);
+  _onTilesetLoaded(tileset) {
+    const root = tileset._root;
+    // TODO: won't be needed after data PR merges, just use root
+    const tileHeader = tileset.isRoyal ? root.children[0] : root;
+    const {center} = tileHeader.boundingVolume;
```
xintongxia: For `sphere` and `region` [cases](https://github.com/uber-web/loaders.gl/blob/master/modules/3d-tiles/src/tileset/helpers/bounding-volume.js#L35), `tileHeader.boundingVolume.center` is in cartographic.

examples/3d-tiles/tile-3d-layer.js
loshjawrence: It's for everything. I can get rid of the +16 and the comment.

examples/3d-tiles/app.js
loshjawrence: Only region does carto at this point. I was testing with royal tilesetpoints and some random pointcloud examples. Not sure if there's an example with a region bounding volume to test with.

examples/3d-tiles/app.js
```diff
@@ -206,30 +215,31 @@ export default class App extends PureComponent {
...
+    if (!center) {
+      // eslint-disable-next-line
+      console.warn('center was not pre-calculated for the root tile');
+    } else {
+      scratchLongLat.copy(center);
```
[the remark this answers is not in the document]
loshjawrence: Do we want to include zoom?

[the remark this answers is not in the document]
ibgreen: > Do we want to include zoom? 

I think that ideally yes, we would calculate a web mercator zoom level that covered the bounding volume, as a "suggestion" to the app.

This would allow apps to always show arbitrary tilesets at the appropriate zoom level after load.

But since it some extra work to calculate it, we can add the `zoom` parameter later, or just hard code a "normal" zoom level for now.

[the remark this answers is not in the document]
loshjawrence: Gotcha, the normal zoom level is what's there now.

loshjawrence:
Merged the new master into this so it uses real cartesian.

loshjawrence:
@ibgreen @xintongxia 
sse traversal seems to be behaving properly now that camera position is totally correct. Got rid of a bug I noticed that was a result of using zoom for ellipsoid height. 

The st helens loading makes more sense now as well. Getting closer to it loads more tiles:
Screenshot_20190703_135121

ibgreen:
@loshjawrence I can't load anymore, crashes in `distanceToTile`

PR #295, opened 2019-07-04
Cleanup old tile._content

loshjawrence:
Adds a tileset3d cache to clean up _content from tiles that weren't traversed this frame. These unloaded tiles are removed from the layer map.

Tested with a budget of 1MB on st helens to trigger the codes paths make sure things behaved as expected.

This is port of the cesium impl.


modules/3d-tiles/src/utils/doubly-linked-list.js
```diff
@@ -0,0 +1,103 @@
...
+import DoublyLinkedListNode from '../utils/doubly-linked-list-node';
+
+const defined = x => x !== undefined;
+
+/**
+ * Doubly linked list
+ *
+ * @private
+ */
+export default class DoublyLinkedList {
```
xintongxia: Maybe can bring the tests for this class later.

PR #316, opened 2019-08-12
Add additional check for jpeg headers for isImage (#315)

twojtasz:
Only the first 2 bytes where being tested. This adds a test for
the 1st byte of the second marker.

Validating the End Of Image would make this more robust, but since
it is expected to test on partial data adding that is not reasonable
at this time.

See the added test case for floating point values that encode as the valid
jpeg header values.


twojtasz:
I verified this fixes the AVS issue reported for now.  Once this is approved I will try to release 1.1.7

Issue #340, opened 2019-08-21
Cannot read property 'match' of undefined


BH-M87:
I met this problem, it caused by sometimes the process is not undefined but the process.version is undefined, in @loaders.gl/core/dist/esm/utils/globals.js:45, `process.version.match` will cause this problem.
Cannot read property 'match' of undefined
[...]

PR #341, opened 2019-08-21
fix bug with read property match of undefined (#340)

BH-M87:
Related to issue #340.


BH-M87:
May I ask when next version will release? This issue block all apps, it can not start at all.

PR #382, opened 2019-09-05
integrate with deck.gl@7.3.0-alpha.6

jianhuang01:
update animationProps to new timeline API


jianhuang01:
> Thanks for updating.
> 
> Clearly not good to have two versions of the Tile3DLayer, that we need to keep in sync...
> 
> To stay sane, as soon as you land your PR in deck and publish a new version, please open Pr that removes `Tile3DLayer` from loaders.gl and imports it from deck.gl.

Of course, will do that after the deck.gl version merged

PR #409, opened 2019-09-17
Fix full uri

xintongxia:
`baseUri` is missing, so the tiles' `fullUri` ended up something like  

`undefined/0/0.b3dm`.


loshjawrence:
Sorry just saw this. The previous PR didn't have this change. I think without this change you could  have multiple nodes in the tree who have the key `/` or just the _basePath because they are empty, but I'm not sure. If that's the case, we don't render empty tiles so it's ok that their keys collide. I think this is ok to include for constant tile initialization but haven't tested all the existing tilesets. _basePath is fed into other code paths like the `get uri` and `get url`

PR #470, opened 2019-10-03
3d-tiles: Cesium example cleanup

ibgreen:


xintongxia:
drop the debugging code

ibgreen:
> drop the debugging code

Whoops, done

PR #486, opened 2019-10-09
Combined loaders

ibgreen:


xintongxia:
It would be nice if provide a way to use local worker by `yarn start-local-worker`

PR #502, opened 2019-10-13
Add b3dm loading to CesiumJS example

OmarShehata:
This PR extends the CesiumJS example to support loading of b3dm tiles. It also adds a "side by side" mode that makes it easier to compare what a tileset looks like loaded with CesiumJS's built-in loading & traversal vs Loaders.gl handling all the loading and traversal, and just passing the model (or point cloud) and its location for CesiumJS to render.

Some notes:

* I added in a `setTileVisible` in `tileset-loader.js` that can be the callback for when Loaders.gl emits an event when any given tile's visibility changes
* I noticed with B3dm tilesets, it seems as though 
* I also noticed the `onTileLoad`, fires both for the b3dm/pnts content, but also for the `json` in case of a tileset that links to another tileset. Is it on me as the user of loaders.gl to run `new Tileset3D()` on that tile content, or is that already handled? Right now I'm just ignoring those tiles.
  * This may be related to the above, but I noticed that the b3dm tilesets I'm loading don't seem to unload the parent tiles. So you can see the low res root tile poking through the high res tiles that load in.
* It was really convenient to have the `load` function from `'@loaders.gl/core'`. I was fetching the tile content directly for the point cloud, but the b3dm one needs to be decoded to get the RTC and the glTF buffer, so it's great that Loaders.gl gives me exactly what I need and I don't have to worry about that.

### DIFF
--- a/examples/cesium/3d-tiles/app.js
+++ b/examples/cesium/3d-tiles/app.js
@@ -1,128 +1,74 @@
-/* global Cesium, fetch */
-
-import {Vector3} from 'math.gl';
-import {registerLoaders, setLoaderOptions} from '@loaders.gl/core';
-import {DracoLoader} from '@loaders.gl/draco';
-import {Tileset3D, _getIonTilesetMetadata} from '@loaders.gl/3d-tiles';
-import {Plane} from '@math.gl/culling';
-import {loadBatchedModelTile, loadPointTile} from './tile-parsers';
-
-// set up loaders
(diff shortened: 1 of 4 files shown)


OmarShehata:
Thanks @ibgreen ! I added some bullets to the issue here https://github.com/uber-web/loaders.gl/issues/481 and can update it as I run into or think of other things.

Issue #478, opened 2019-10-07
Add cesium 3D tile spec support to tiling tool


jianhuang01:
We need support 3D tile tiling feature on top of Potree converter, which will write Potree tiles in cesium 3D tile spec.
- [x] setup xcode project to build converter and dependency libs
- [x] create a hierarchy tree from potree binary data
- [x] create tileset json file in cesium spec
- [x] write binary tile data in cesium spec

PR #508, opened 2019-10-18
Add cesium point cloud 3D tile format support

jianhuang01:
for #478 
- initial implementation of the tilling tool which supports cesium 3D spec.
- add some improvements to handle app parameters better.


xintongxia:
Do you have a screenshot?

jianhuang01:
> Do you have a screenshot?

We didn't support georeference for now, when I am using cesium renderer, the data only show up in the first second, then covered by meshes of the earth. Because of the rendering issue, I can't get a screenshot. But I could see the point cloud is rendered correctly.

PR #509, opened 2019-10-20
Hide tiles that are not visible this frame in Cesium example

OmarShehata:
@ibgreen as per our discussion on https://github.com/uber-web/loaders.gl/issues/481#issuecomment-541984920, I updated the example to only show tiles that are visible this frame. But `tileset3d.update` doesn't seem to return anything. The [docs say](https://loaders.gl/modules/3d-tiles/docs) to use `tileset3d.traverse()` but I think that was renamed to update? 

The good news is that I am able to reproduce the tiles disappearing issue we were seeing from before when testing in Deck.gl. There seem to be 2 issues which may be related:

* If you try the Cairo tileset, you'll see it refines to the children, but then as you zoom closer, it seems to traverse backwards, going back to the root tile, before tiles start to disappear.
* In the Malalison island tileset, the issue is not as bad, but you can sometimes see tiles disappearing around the edges (although that may just be culling too early? I wasn't able to get a tile in dead center of the screen to disappear in the way I could with the Cairo tileset, so the issue may be limited to tilesets with external tilesets).

### DIFF
--- a/examples/cesium/3d-tiles/app.js
+++ b/examples/cesium/3d-tiles/app.js
@@ -16,8 +16,8 @@ const VIEW_TYPES = {SINGLE: 0, SIDE_BY_SIDE: 1};
 
 // Application settings.
 const VIEW_MODE = VIEW_TYPES.SIDE_BY_SIDE;
-const TOKEN = Cesium.Ion.defaultAccessToken;
-const ASSET_ID = CAIRO_ASSET_ID;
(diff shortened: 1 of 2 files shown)


xintongxia:
@loshjawrence 

Here is the screenshot, left is with loaders.gl, right is with Cesium.

Screen Shot 2019-11-07 at 12 57 23 PM

PR #534, opened 2019-11-19
fix: change color Uint8Array to Float32Array

zt123123:
before change color array type
my program not work as my expect
68575594-575d4000-04a7-11ea-848f-e039dcd67fc6

I just change the Uint8Array to Float32Array, it works well!
69114607-e4645280-0ac0-11ea-863c-48398a8d2aba


xintongxia:
See https://github.com/uber-web/loaders.gl/pull/535

PR #543, opened 2019-12-05
deck.gl v8.0 API audit

xintongxia:
v8.0 API audit


Pessimistress:
We are not making a major release of loaders. Can we not make a breaking change now?

PR #557, opened 2019-12-12
Add i3s example

xintongxia:


xintongxia:
> Do we want separate examples for each 3D tile format rather than a dropdown? Lot of duplicated code...

Based on luma and deck examples/demo pages, it list each example on its own, then poeple would have a direct idea what are inside loaders.gl instead of hiding all of them in one place.

PR #579, opened 2019-12-21
Port getFrameState from deck and update Tileset3D.update API

xintongxia:
- Port getFrameState from deck
- Tileset3D.update(frameState) => Tileset3D.update(viewport) : frameNumber
- deleted local version of Tile3DLayer, instead import from `@deck.gl/geo-layers`


modules/3d-tiles/src/lib/tileset/helpers/get-frame-state.js
```diff
@@ -18,7 +18,7 @@ const cullingVolume = new CullingVolume([
 export function getFrameState(viewport, frameNumber) {
   // Traverse and and request. Update _selectedTiles so that we know what to render.
   const {cameraDirection, cameraUp, height} = viewport;
-  const {metersPerPixel} = viewport.distanceScales;
+  const {metersPerUnit} = viewport.distanceScales;
 
   const viewportCenterCartographic = [viewport.longitude, viewport.latitude, 0];
```
Pessimistress: Not for this PR: now that we are using `viewport` (aka deck.gl `WebMercatorViewport`) as an argument, we should reduce the API surface between the two modules.

Here we are using both `viewportCenterCartographic = [viewport.longitude, viewport.latitude, 0]` and `viewportCenterCommonSpace = viewport.center`. Ideally we'll choose one and use `projectPosition`/`unprojectPosition` to convert between them. Technically those two are not the same location (center may have z).

modules/3d-tiles/src/lib/tileset/tileset-3d.js
```diff
@@ -259,7 +261,18 @@ export default class Tileset3D {
...
-  update(frameState) {
+  update(viewport) {
+    let frameState;
+    if ('frameNumber' in viewport) {
+      // backward compatibility
+      // this is using old API, input is `frameState` object
+      // old API: update(frameState)
+      frameState = viewport;
+    } else {
+      // TODO deprecated in v8.x
```
Pessimistress: It should be the other one that's deprecated?

PR #627, opened 2020-01-15
@loaders.gl/3d-tiles: Port DoublyLinkedList test from Cesium

xintongxia:
- port test file
- fix linked list counter


xintongxia:
> * A doubly linked list does not seem to be among the most common data structure in JS - I am not even sure I have seen a JS implementation before.
> * Looking forward to see your use case for this class.

It has been used in Tile3DCache and thus part of the Traversal.

PR #631, opened 2020-01-22
3d-tiles and i3s: Add RFC for consolidating 3d tiles decoding

xintongxia:
TODO

- Part of links in the docs are not valid


dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,99 @@
...
+- [OGC 3D Tiles](https://www.opengeospatial.org/standards/3DTiles) standard
+- [OGC i3s](https://www.opengeospatial.org/standards/i3s) standard
+- **Tile Header Hierarchy** - An initial, "minimal" set of data listing the _hierarchy of available tiles_, with minimal information to allow an application to determine which tiles need to be loaded based on a certain viewing position in 3d space.
+- **Tile Header** - A minimal header describing a tiles bounding volume and a screen space error tolerance (allowing the tile to be culled if it is distant), as well as the URL to load the tile's actual content from.
+- **Tile Cache** - Since the number of tiles in big tilesets often exceed what can be loaded into available memory, it is important to have a system that releases no-longer visible tiles from memory.
+- **Tileset Traversal** - Dynamically loading and rendering 3D tiles based on current viewing position, possibly triggering loads of new tiles and unloading of older, no-longer visible tiles.
+
+## Components
+
+- [`Tileset3DLoader`](docs/api-reference/3d-tiles/tileset-3d-loader): Load a tileset into a json object.
```
Pessimistress: The links do not work

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,99 @@
...
+});
+
+// load i3s tileset
+const tilesetJSON = await load(tilesetUrl, I3STilesetLoader);
+```
+
+2. Construct `Tileset3D` instance
+
+```js
+const tileset3d = new Tileset3D(tilesetJson, tilesetUrl, {
```
Pessimistress: Why is `tilesetUrl` needed again? Should it be part of the `tilesetJSON`?

modules/3d-tiles/docs/api-reference/tile-3d-header.md
```diff
@@ -0,0 +1,124 @@
+# Tile3DHeader
```
Pessimistress: Is this a proposal or does this match the current master?

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,99 @@
...
+const tilesetUrl = 'A tileset url';
+
+// load 3d tiles from ion
+const tilesetJSON = await load(tilesetUrl, Tileset3DLoader, {
+  ionAssetId,
+  ionAssetToken
+});
+
+// load i3s tileset
+const tilesetJSON = await load(tilesetUrl, I3STilesetLoader);
```
Pessimistress: `I3STilesetLoader` is not imported

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,99 @@
...
+## Concepts
+
+- [OGC 3D Tiles](https://www.opengeospatial.org/standards/3DTiles) standard
+- [OGC i3s](https://www.opengeospatial.org/standards/i3s) standard
+- **Tile Header Hierarchy** - An initial, "minimal" set of data listing the _hierarchy of available tiles_, with minimal information to allow an application to determine which tiles need to be loaded based on a certain viewing position in 3d space.
+- **Tile Header** - A minimal header describing a tiles bounding volume and a screen space error tolerance (allowing the tile to be culled if it is distant), as well as the URL to load the tile's actual content from.
+- **Tile Cache** - Since the number of tiles in big tilesets often exceed what can be loaded into available memory, it is important to have a system that releases no-longer visible tiles from memory.
+- **Tileset Traversal** - Dynamically loading and rendering 3D tiles based on current viewing position, possibly triggering loads of new tiles and unloading of older, no-longer visible tiles.
+
+## Components
```
Pessimistress: Start with answering the following questions:

- How many modules will there be?
- What are the exports of each module? What formats do they load to?
- Do they share any common components?

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. Load a tileset
+
+```js
+import {Tileset3DLoader, Tileset3D} from '@loaders.gl/3d-tiles';
+import {load} from '@loaders.gl/core';
+
+const tilesetUrl = 'A tileset url';
+
+// load 3d tiles from ion
```
twojtasz: Maybe `3DTilesTileSetLoader` since the idea is various 3d tile formats could be supported and it would be similar to `I3STileSetLoader` in the naming convention

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,103 @@
...
+const tilesetJSON = await load(tilesetUrl, Tileset3DLoader, {
+  ionAssetId,
+  ionAssetToken
+});
+
+// load i3s tileset
+const tilesetJSON = await load(tilesetUrl, I3STilesetLoader);
+```
+
+2. Construct `Tileset3D` instance
```
twojtasz: So the loader provides the **Tile Header Hierarchy**, this is passed in when constructing the TileSet3d object,  which uses the heirarchy to actually load the tiles based on the viewport.  Just making sure I understand what data is actually used and download in each class

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,103 @@
...
+      default:
+        throw new Error(`Tile3DLayer: Failed to render layer of type ${tileHeader.layerType}`);
+    }
+  }
+});
+```
+
+## Questions
+
+It is unrealistic to support all the details for both SPECs. What is a reasonable goal for the 3d tiles loaders?
```
twojtasz: Do we have a list of features that are incompatible?

dev-docs/RFCs/v2.1/normalize-3d-tile-loaders.md
```diff
@@ -0,0 +1,396 @@
...
+
+**For shared by tiles loaders**
+
+- `Tile2D`: A class that extended from 2D tile data.
+- `Tile3D`: A class that can understand unified tile format loaded from tile loaders, i.e. `Tiles3DTileLoader` for `@loaders.gl/3d-tiles`, `I3STileLoader` for `@loaders.gl/i3s`.
+- `TilesetTraversal`: A class that traverse tile tree to get renderable tiles based on certain state (i.e. viewport), may cause loading and unloading tiles, and updating tiles visibilties.
+- `TilesetCache`: LRU cache for caching loaded tiles.
+- `RequestScheduler`: Managing fetching requests based on traversal priorities.
+
+```
```
Pessimistress: How does `Tileset3D` know which loaders to use to load the tile contents? Per module dependency hierarchy described above, the `@loaders.gl/tiles` module should not depend on any spec implementations. Is the user supposed to register the `*TileLoader` and `*TileContentLoader` as well? This is not apparent from the code sample in this RFC.

If the usage of `*TileLoader` and `*TileContentLoader` are hidden from the user, does it still make sense to have so many loaders? It is possible for a single loader to handle multiple file formats. The parser can make the switch based on url, as well as content sniffing. It will be a lot less confusing if each module only exports one loader, e.g. `Tiles3DLoader` or `I3SLoader`.

xintongxia: I agree that the multiple loaders could be confusing to the user. But normally the tile header contains metadata and a `url` pointing to the actual payload, which is the `tile.content` in this proposal. 

Another way, each loader could expose 
- a `TileLoader` which accept options: `loadContent`
- a parser to parse the content

PR #657, opened 2020-02-25
Fix travis build by skip downloading Chromium

xintongxia:


xintongxia:
> * Is Chromium not needed to run the browser mode tests?
> * On our local machines we have a global install, so we can skip it, but I assume that is not true on Travis?
> * If this speeds up, should we set this in all our repos?

I just tried to fix the build failures, the error told me to skip downloading. But apparently it didn't fix the issues.

PR #665, opened 2020-03-05
[tiles]: fix traversal - select root node when it is visible

xintongxia:
root tile should be selected for render when it is visible.


modules/tiles/src/tileset/traversers/tileset-traverser.js
```diff
@@ -43,15 +43,8 @@ export default class TilesetTraverser {
     // update tile (visibility and expiration)
     this.updateTile(root, frameState);
 
-    // check if traversable
-    if (!this.canTraverse(root, frameState, true)) {
```
xintongxia: after this removed, whether `root` tile should be selected for rendering will be checked in `executeTraversal`

PR #678, opened 2020-03-10
images: additional jpeg check

kvalev:
As an additional safeguard for false positives, check the end of image marker [wikipedia](https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format#File_format_structure).


xintongxia:
@kvalev thanks for your contributions! Do you mind adding the wiki link to the code?

PR #680, opened 2020-03-11
[3d-tiles]: Add CesiumIonLoader to allow resolving authorization in preload

xintongxia:
For #676

### DIFF
--- a/examples/deck.gl/3d-tiles/app.js
+++ b/examples/deck.gl/3d-tiles/app.js
@@ -62,7 +62,7 @@ export default class App extends PureComponent {
       // EXAMPLE STATE
       droppedFile: null,
       examplesByCategory: null,
-      selectedExample: {},
+      selectedExample: null,
       category: INITIAL_EXAMPLE_CATEGORY,
       name: INITIAL_EXAMPLE_NAME
     };
--- a/examples/deck.gl/3d-tiles/app.js
+++ b/examples/deck.gl/3d-tiles/app.js
@@ -142,7 +142,7 @@ export default class App extends PureComponent {
   }
 
   // Called by ControlPanel when user selects a new example
-  async _onSelectExample({example, category, name}) {
+  _onSelectExample({example, category, name}) {
     this.setState({selectedExample: example, category, name});
   }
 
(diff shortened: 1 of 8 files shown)


examples/deck.gl/3d-tiles/app.js
```diff
@@ -108,14 +110,7 @@ export default class App extends PureComponent {
...
   async _loadExample(selectedExample) {
-    const {ionAssetId, ionAccessToken} = selectedExample;
-    if (ionAssetId && ionAccessToken) {
-      const metadata = await _getIonTilesetMetadata(ionAccessToken, ionAssetId);
-      const {url, headers} = metadata;
-      selectedExample.tilesetUrl = url;
-      selectedExample.headers = headers;
-      selectedExample.metadata = metadata;
-    }
+    selectedExample.tilesetUrl = `${TILESET_SERVER_URL}/${selectedExample.ionAssetId}/tileset.json`;
```
Pessimistress: - This is no longer async
- Don't mutate constants - just construct this URL in `render`

examples/deck.gl/3d-tiles/tile-3d-layer.js
```diff
@@ -69,13 +69,17 @@ export default class Tile3DLayer extends CompositeLayer {
 
   async _loadTileset(tilesetUrl) {
```
Pessimistress: We should remove this class from examples and use deck.gl's Tile3DLayer

modules/3d-tiles/src/index.js
```diff
@@ -1,5 +1,6 @@
 // LOADERS
 export {default as Tiles3DLoader} from './tiles-3d-loader';
+export {default as IONLoader} from './ion-loader';
```
Pessimistress: `CesiumIonLoader`?

modules/3d-tiles/src/ion-loader.js
```diff
@@ -0,0 +1,18 @@
+import Tiles3DLoader from './tiles-3d-loader';
+import {getIonTilesetMetadata} from './lib/ion/ion';
+
+async function preload(url, options) {
```
Pessimistress: `options` is a nested object for all loader modules: `image`, `fetch`, etc.

examples/deck.gl/3d-tiles/tile-3d-layer.js
xintongxia: Yes, I always keep them sync and will delete it before this PR is landed.

modules/3d-tiles/src/ion-loader.js
xintongxia: The nested options is tricky, that means I need copy options['cesium-ion'] to/from options['3d-tiles'].

PR #705, opened 2020-03-24
kml: fix parsing coordinate numbers

xintongxia:
the parsed `coordinates` should be numbers instead of strings.


xintongxia:
@tylerprete

PR #719, opened 2020-04-10
Fix imports in JSONLoader docs

kylebarron:
The streaming examples in JSON docs had incorrect imports.

Additionally, should `loadInBatches` be listed as a Supported API? Right now only `load` is listed in the table, which might be confusing?

| Loader         | Characteristic                                       |
| -------------- | ---------------------------------------------------- |
| Supported APIs | `load`, `parse`, `parseSync`, `parseInBatches`       |


xintongxia:
@kylebarron 

Thanks for the fix! But the CI is still pending, could you rebase against master?

PR #736, opened 2020-05-07
tiles: fix loading tiles after resuming traversal

xintongxia:
Problem:

The Tileset loading and rendering flow is as below,
`Tile3DLayer` triggers `tileset.update()` after a `tileset` is loaded
`tileset` starts traversing the tile tree and update `requestTiles` (tiles of which content need to be fetched) and `selectedTiles` (tiles ready for rendering under the current viewport). `Tile3DLayer` will update rendering based on `selectedTiles`. `Tile3DLayer` also listens to `onTileLoad` callback and trigger another round of `update/traversal` when new tiles are loaded.

As I3S tileset have stored `tileHeader` file (metadata) and tile content files (geometry, texture, ...) separately. During each traversal, it issues `tilHeader` requests if that `tileHeader` is not yet fetched, after the tile header is fulfilled, it will resume the traversal starting from the tile just fetched.
 
The original code didn't update `requestedTiles` or `selectedTiles` after each resumed traversal, hence some tiles are not being rendered as expected.

```
Tile3DLayer
     |
 await load(tileset)
     |
 tileset.update()
     |                async load tileHeader 
 tileset.traverse() -------------------------- Queued
     |        resume traversal after fetched  |
     |----------------------------------------|
     | 
     |                     async load tile content
tilset.requestedTiles  ----------------------------- RequestScheduler                                                                
                                                            |
tilset.selectedTiles (ready for rendering)                  |
     |                                                      |
  Tile3DLayer    Tile3DLayer -------------------------------|
     |                 |      notify new tile is available       
  updateLayers       onTileLoad
                        |
                      tileset.update // trigger another round of update
```


xintongxia:
> I am a bit wary about this approach, i.e. having the traversal operation being considered async.
> 
> The reason is that applications may request traversal multiple times for other reasons, and this could potentially result in multiple outstanding traversal calls, which could lead to a number of issues.
> 
> In the initial implementation for 3D tiles that I put up, traversal was always sync - it returned any loaded tiles and scheduled loads of any other tiles, but did so asynchronously.
> 
> The idea was that whenever a tile loads, the app traverses again, getting a bigger list of loaded tiles, until all tiles in view are loaded.
> 
> That said, the changes in this PR make sense based on the chosen strategy.

I understand your concerns about one `update` call triggering multiple traversals. Async traversal is causing troubles. As the `traverser` is not able to continue without knowing the tile nodes' level of detail error, which results in fetching-then-resume-traversal loop. 

Traversal in current approach is kind of async, but
- Tile metadata file will be only fetched once.
- Upon tile metadata fetched, the traversal will be only resumed when the tile is still needed in current frame state (or viewport).

Also I'll sync with Tam to see how can we take use of `nodePage` file (based on v1.7 spec, the `nodePage` files contains metadata of all the nodes)

FYI, for `3D tiles` case, it sticks to one `update` call will trigger exactly one `traversal`.

PR #741, opened 2020-05-14
Fix for Node.js being unable to pull files from non local FS

NikolayActionEngine:
Node.js polyfill wrapper (modules/polyfills/src/fetch-node/read-file.node.js) wasn't closing request making it unable to complete for a non-local FS resources.

Fixed this as well as made test cases for node to pull resources from github directly instead of a local file system. (those 2 were failing before the change if pointed to http/https resources) and passing now.


kylebarron:
I believe I've been hitting this error as well, but hadn't taken the time to debug.

PR #781, opened 2020-06-12
Experimental TARBuilder

chrisgervang:
Derived from tar-js


chrisgervang:
Agree with all comments and will implement them in a follow up PR, since they are not user facing.

PR #799, opened 2020-06-17
wip streaming shp parser

ibgreen:


ibgreen:
Will reopen once we get serious about streaming

PR #916, opened 2020-08-21
Shapefile: Parse SHP/DBF in batches

kylebarron:
- I haven't figured out yet how to integrate batched parsing into the top-level `ShapefileLoader`. It seems I need to set a specific `batchSize` so that the returned geometries and attributes represent the same number of rows?
- There are some SHP tests already for loading in batches, but more fine-grained tests seem ideal


kylebarron:
Superseded by #925

PR #953, opened 2020-09-23
Set worker url to local file for shapefile tests

kylebarron:
Tests were unreliable/failing because shapefile tests were trying to use remote worker bundles.


Pessimistress (approved):

ibgreen (approved):
Just verify there are no other loader.spec.js fies in this module that needs the same treatment...

kylebarron:
Tests passed locally only adding it to the shapefile tests, but it doesn't hurt to also add it to the two others.

PR #964, opened 2020-10-08
Added attributes loader for I3s.

dryabinin-actionengine:


modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+
+  for (let index = 0; index < items.length; index++) {
+    const fileOrUrl = items[index];
+
+    switch (type) {
+      case FILES:
+        promises.push(fetchFile(fileOrUrl));
+        break;
+      case ARRAY_BUFFERS:
+        promises.push(fileOrUrl.arrayBuffer());
```
ibgreen, line 37: This switch statement seems a bit weird to me. 
- The urls and files are very different.
-  It feels like this should be split into two functions, even though some iteration is duplicated. 
- Also make sure to add a JSDoc line above each function.

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+        break;
+      default:
```
ibgreen, line 39: This is an error if you split the function you get rid of this case.

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+        break;
+    }
+  }
+
+  return await Promise.all(promises);
+}
+
+function getTileAttributes(arrayBuffers, options) {
```
ibgreen, line 47: JSDoc

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+  const attributes = {};
+  const {attributeStorageInfo} = options;
```
ibgreen, line 49: It is a bit strange to have something as important as `attributeStorageInfo` in a parameter named `options`.  I mean this code doesn't even work without attributeStorageInfo, so why not just make that the parameter?

`options` is usually our loader options passed down.

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+  for (let index = 0; index < attributeStorageInfo.length; index++) {
+    const attributeName = attributeStorageInfo[index].name;
+    const attributeType = attributeStorageInfo[index].attributeValues.valueType;
+    attributes[attributeName] = parseAttributes(attributeType, arrayBuffers[index]);
+  }
+
+  return attributes;
+}
+
+function parseAttributes(attributeType, arrayBuffer) {
```
ibgreen, line 60: JSDOc

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+    case OBJECT_ID_ATTRIBUTE_TYPE:
+      return parseShortNumberAttributes(arrayBuffer);
+    case FLOAT_64_TYPE:
+      return parseFloatAttributes(arrayBuffer);
+    default:
+      return parseShortNumberAttributes(arrayBuffer);
+  }
+}
+
+function parseShortNumberAttributes(arrayBuffer) {
```
ibgreen, line 73: Ditto

modules/i3s/src/i3s-attributes-loader.js
```diff
@@ -0,0 +1,119 @@
...
+    const stringAttribute = Buffer.from(arrayBuffer, stringOffset, stringByteSize).toString();
+    stringsArray.push(stringAttribute);
+    stringOffset += stringByteSize;
+  }
+
+  return stringsArray;
+}
+
+/** @type {LoaderObject} */
+const I3SAttributesLoader = {
```
ibgreen, line 104: You might want to keep this file relatively small focusing on the LoaderObject and move the attribute parsing functions to the `src/lib` directory.

modules/i3s/test/i3s-attributes-loader.spec.js
```diff
@@ -0,0 +1,92 @@
+import test from 'tape-promise/tape';
+import I3SAttributesLoader from '../src/i3s-attributes-loader';
+
+const attributeStorageInfo = [
```
ibgreen, line 4: +100, love the tests, thank you!!!

ibgreen (approved):
This looks good. I assume some of this attribute loading is already happening in the I3SLoader (after all we are able to render I3S tiles today).

So I assume the next PR will replace existing code with this loader?

It could be useful if you explain such context / plans in the PR description.

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: Done

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: Done

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: Replaced `options` -> `attributeStorageInfo`

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: done

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: done

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: moved to another file in `src/lib`

modules/i3s/test/i3s-attributes-loader.spec.js
dryabinin-actionengine: > This looks good. I assume some of this attribute loading is already happening in the I3SLoader (after all we are able to render I3S tiles today).
> 
> So I assume the next PR will replace existing code with this loader?
> 
> It could be useful if you explain such context / plans in the PR description.

There are so many `attributes` words related to `I3SLoader`.
Already existing attribute loading functionality in `I3SLoader` is responsible for loading `geometry attributes` such as:
image

My new loader is responsible for `feature attributes` which is not the same as previous.
This attributes characterize particular feature (building or some object on the map).
For example:
image

modules/i3s/src/i3s-attributes-loader.js
dryabinin-actionengine: Removed it and split on two functions.

PR #975, opened 2020-10-14
i3s: add minor comment

belom88:


ibgreen (approved):
You could add the explanation that it includes data from multiple files.

PR #1002, opened 2020-10-30
Change Draco modules to be created asynchronously

FrankGalligan:
In general Draco npm modules have always been created asynchronously.

We are planning to release Draco 1.4.0, which uses a new version of emscripten and causes applications to fail more often if they were getting created synchronously as the timing has changed.

Will this change work in your code?

### DIFF
--- a/modules/draco/src/lib/draco-module-loader.js
+++ b/modules/draco/src/lib/draco-module-loader.js
@@ -15,10 +15,8 @@ export async function loadDracoDecoderModule(options) {
     loadDecoderPromise =
       loadDecoderPromise ||
       new Promise(resolve => {
-        const draco = modules.draco3d.createDecoderModule({
-          onModuleLoaded() {
-            resolve({draco});
-          }
+        modules.draco3d.createDecoderModule({}).then(function(module) {
+          resolve(module);
         });
       });
   } else {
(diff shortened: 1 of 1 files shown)


belom88 (changes requested):
Tests hang up on this branch.
@FrankGalligan , notice that loaders now handle 1.3.6 draco version (see https://github.com/visgl/loaders.gl/blob/f1c89f1907c474436275ae8fd451d1fb315814b7/modules/draco/package.json#L44)
Should 1.3.6 handle your changes? If not - we need to update loaders to 1.4.0 first.

FrankGalligan:
I think so but I wasn't sure how to check in this project.

ibgreen:
Just testedm this seems to run fine on Node.js and on browser main thread.

But the `DracoWorkerLoader` hangs. I did rebuild it using `yarn bootstrap` before testing. 

@FrankGalligan not sure if that is a clue, if any consideration using this callback mode on a worker thread?

Screen Shot 2020-11-12 at 4 23 53 PM

belom88:
@FrankGalligan . Thank you for this PR. We used you approach in https://github.com/visgl/loaders.gl/pull/1029 . This PR can be closed

PR #1003, opened 2020-11-02
loader-utils: new features

belom88:
- concatenateTypedArrays function;
- padToNBytes function;
- ChildProcessProxy - wait === 0 - infinity


ibgreen (approved):
- Remove error strings that add to the bundle size.
- Remove `pad4` and replace all use with `padN`

PR #1042, opened 2020-12-10
Fixed base map selection for examples in  I3S and 3DTiles

dryabinin-actionengine:


ibgreen (approved):
I see the temptation of sharing constants between examples. One design constraint is that we wanted the user to be able to copy each example folder standalone outside of the repo and run it.

Other than that, looks good.

dryabinin-actionengine:
> I see the temptation of sharing constants between examples. One design constraint is that we wanted the user to be able to copy each example folder standalone outside of the repo and run it.
> 
> Other than that, looks good.

Fixed. Split into two files

ibgreen (approved):

PR #1047, opened 2020-12-11
Added feature ids support for compressed geometry 3DTiles -> I3s converter

dryabinin-actionengine:
1) Features in compressed geometry becomes available
2) Fixed color of tileset without textures (New York)


ibgreen (approved):

Tamrat-B:
@dryabinin-actionengine that's great news. I'll give it a shot.
thx

PR #1056, opened 2020-12-15
draco: get rid of local libs

belom88:
It fixes https://github.com/visgl/loaders.gl/issues/1051 but now "draco" module gets the encode library from google github: https://raw.githubusercontent.com/google/draco/${DRACO_VERSION}/javascript/draco_encoder.js


ibgreen (approved):
Good that we have a fix. But we do know that some users are in air-gapped environments and cannot load scripts from the internet. They need an option to bundle or load the scripts from local env, so this will become an issue again.

belom88:
> Good that we have a fix. But we do know that some users are in air-gapped environments and cannot load scripts from the internet. They need an option to bundle or load the scripts from local env, so this will become an issue again.

They still can install npm package 'draco3d' and use it as module

PR #1061, opened 2020-12-17
draco: increase decompression performance

belom88:
Current master:
Screenshot from 2020-12-17 06-32-38
This branch:
Screenshot from 2020-12-17 06-29-51

I checked this on local loader (not worker).

### DIFF
--- a/modules/draco/src/lib/draco-parser.js
+++ b/modules/draco/src/lib/draco-parser.js
@@ -217,19 +217,14 @@ export default class DracoParser {
   _getMeshFaceIndices(decoder, dracoGeometry) {
     // Example on how to retrieve mesh and attributes.
     const numFaces = dracoGeometry.num_faces();
-
     const numIndices = numFaces * 3;
-    const indices = new Uint32Array(numIndices);
-    const dracoArray = new this.draco.DracoInt32Array();
-    for (let i = 0; i < numFaces; ++i) {
-      decoder.GetFaceFromMesh(dracoGeometry, i, dracoArray);
-      const index = i * 3;
-      indices[index] = dracoArray.GetValue(0);
-      indices[index + 1] = dracoArray.GetValue(1);
-      indices[index + 2] = dracoArray.GetValue(2);
-    }
+    const byteLength = numIndices * 4; // TODO: 4 - is type size of indices array. Possibly this data is defined in header
 
-    this.draco.destroy(dracoArray);
+    const ptr = this.draco._malloc(byteLength);
+    decoder.GetTrianglesUInt32Array(dracoGeometry, byteLength, ptr);
+    const indices = new Uint32Array(this.draco.HEAPF32.buffer, ptr, numIndices).slice(); // TODO: why slice here?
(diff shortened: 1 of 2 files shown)


ibgreen:
This is very impressive - surprisingly so, as per the THREE.js issue I got the sense that the improvement should be around 20%, not 100%.

Any chance you can also test workers using the throughput measurement options we use for images?

There should be a way to reduce the number of iterations when setting up the Bench suite to make the benchmarks run faster.

modules/draco/src/lib/draco-parser.js
```diff
@@ -217,19 +219,14 @@ export default class DracoParser {
...
-      const index = i * 3;
-      indices[index] = dracoArray.GetValue(0);
-      indices[index + 1] = dracoArray.GetValue(1);
-      indices[index + 2] = dracoArray.GetValue(2);
-    }
+    const byteLength = numIndices * INDEX_ITEM_SIZE;
+
+    const ptr = this.draco._malloc(byteLength);
+    decoder.GetTrianglesUInt32Array(dracoGeometry, byteLength, ptr);
+    const indices = new Uint32Array(this.draco.HEAPF32.buffer, ptr, numIndices).slice();
```
ibgreen, line 227: Why do we `slice`? That creates a copy?

modules/draco/src/lib/draco-parser.js
```diff
@@ -263,71 +260,54 @@ export default class DracoParser {
...
-        break;
+    const ptr = this.draco._malloc(byteLength);
+    decoder.GetAttributeDataArrayForAllPoints(
+      dracoGeometry,
+      dracoAttribute,
+      dataType,
+      byteLength,
+      ptr
+    );
+    const typedArray = new TypedArrayCtor(this.draco.HEAPF32.buffer, ptr, numValues).slice();
```
ibgreen, line 279: Slice seems to create unnecessary copy?

ibgreen (approved):
Would be good to do some visual tests too. 
- Temporarily re-enable the render tests
- Test some DRACO compressed glTF models in `examples/website/gltf` (using `yarn start-local`)
- Test some glTF powered 3D Tiles modes in `examples/website/3d-tiles`.
-

belom88:
I've done some testing with local worker: 
```
workerUrl: `modules/draco/dist/draco-loader.worker.dev.js`,
```
Current master:
Screenshot from 2020-12-18 14-41-44
This branch:
Screenshot from 2020-12-18 14-44-59
All visual tests pass.

modules/draco/src/lib/draco-parser.js
belom88: Wrapped `this.draco.HEAPF32.buffer` have to be disposed. A copy is necessary

modules/draco/src/lib/draco-parser.js
belom88: Wrapped `this.draco.HEAPF32.buffer` have to be disposed. A copy is necessary

PR #1076, opened 2021-01-01
Compressed textures: pvr & dds

belom88:
* PVR & DDS files are loaded;
* Examples don't work for now because there isn't the package "@loaders.gl/texture" in npm. We have a way now to work around this issue and text compressed textures until "texture" module isn't published;
* Basis example doesn't work because of  wrong link to "worker". I expect it will be fixed as soon as "texture" module is published. We can also use {worker: false} option for now.

I plan to merge it with merge commit to save Daniil's and mine commit history.


ibgreen (approved):
Looks good. 

Some test cases around loading and checking of loaded data would be good.

PR #1077, opened 2021-01-01
polyfills: follow redirect option

belom88:


ibgreen (approved):
Is there are good reason to keep this option `false` by default? 

How does fetch work on browser, does it follow redirects? I guess we should emulate that?

belom88:
You are right. Following redirect is default behavior of `fetch` API (instead of NodeJS `http` API).
https://developer.mozilla.org/en-US/docs/Web/API/Response/redirected#Disallowing_redirects

PR #1093, opened 2021-01-13
Move file sizes logic to loaders in textures app

dryabinin-actionengine:


ibgreen (commented):
I see. This is too specific too the image loader. It needs to be more generic.

You could just split the load into a fetch and a parse.

```js
const response = await fetchFile(url, options);
const arrayBuffer = await response.arrayBuffer();
const length = arrayBuffer.byteLength;
const image = load(arrayBuffer, loaders, options);
```

dryabinin-actionengine:
> I see. This is too specific too the image loader. It needs to be more generic.
> 
> You could just split the load into a fetch and a parse.
> 
> ```js
> const response = await fetchFile(url, options);
> const arrayBuffer = await response.arrayBuffer();
> const length = arrayBuffer.byteLength;
> const image = load(arrayBuffer, loaders, options);
> ```

Thanks. Yes it will be better. Changed this.

ibgreen (approved):

PR #1103, opened 2021-01-15
Add 'baseUri' to default options

dryabinin-actionengine:


ibgreen (changes requested):
Let's move this PR to 2.3-release branch, and land it to remove the warning. 

But for master / 3.0, this option should be removed completely in glTF loader, this may require more careful changes.

PR #1121, opened 2021-01-18
Move NPYLoader to textures module

kylebarron:
#### Change List

- Move `NPYLoader` to `textures` module
- Make non-experimental export, i.e. export `NPYLoader` instead of `_NPYLoader`


ibgreen (approved):
Can we add a from 3.0 badge in the NPYLoader doc page?

PR #1134, opened 2021-01-26
Fix for Pre-built bundles broken in 2.3.9

dryabinin-actionengine:
@ibgreen  I tested it to create the same logic is [here](https://codepen.io/Pessimistress/pen/LYRKMqK?editors=1001) but locally.
Also we should implements some test cases for scripts too.

### DIFF
--- a/scripts/bundle.config.js
+++ b/scripts/bundle.config.js
@@ -44,10 +44,10 @@ const NODE = {
 
 const BABEL_CONFIG = {
   presets: [
-    ['@babel/preset-env', {modules: false}]
+    ['@babel/preset-env', {forceAllTransforms: true}]
   ],
   plugins: [
-    ['@babel/plugin-transform-runtime', {useESModules: false}], 
+    '@babel/transform-runtime',
     'version-inline'
   ]
 };
--- a/modules/csv/src/libs/papaparse.js
+++ b/modules/csv/src/libs/papaparse.js
@@ -37,7 +37,7 @@ var global = (function() {
 var IS_PAPA_WORKER = false;
 
 var Papa = {};
-module.exports = Papa;
+export default Papa;
 Papa.parse = CsvToJson;
 Papa.unparse = JsonToCsv;
 
--- a/modules/las/src/libs/laz-perf.js
+++ b/modules/las/src/libs/laz-perf.js
@@ -23,7 +23,7 @@
  */
 
 // laz-perf.js
-module.exports = function getModule() {
+export default function getModule() {
   var Module = typeof Module !== "undefined" ? Module : {};
   var moduleOverrides = {};
   var key;
--- a/modules/polyfills/src/libs/encoding.js
+++ b/modules/polyfills/src/libs/encoding.js
@@ -3081,4 +3081,4 @@ decoders['x-user-defined'] = function(options) {
 // if (!global['TextEncoder']) global['TextEncoder'] = TextEncoder;
 // if (!global['TextDecoder']) global['TextDecoder'] = TextDecoder;
 // babel.config.js skip transpiling files in `libs/`
-module.exports = {TextEncoder, TextDecoder};
+export {TextEncoder, TextDecoder};


scripts/bundle.config.js
```diff
@@ -49,7 +49,9 @@ const BABEL_CONFIG = {
   plugins: [
     ['@babel/plugin-transform-runtime', {useESModules: false}], 
     'version-inline'
-  ]
+  ],
+  // Consider the file a "module" if import/export statements are present, or else consider it a "script".
+  sourceType: 'unambiguous'
```
ibgreen, line 54: How does this help us? What is "the file" that is being considered a module here?

We do have a couple of script based examples do they fail to load without this change, and succeed to load with this change?

Pessimistress: Why is `modules:false` set? I don't think you want to preserve ES modules: https://babeljs.io/docs/en/babel-preset-env#modules

deck.gl's bundle settings: https://github.com/visgl/deck.gl/blob/master/scripts/bundle.config.js#L65

dryabinin-actionengine: @ibgreen @Pessimistress 

The problem - `dist.min.js:1 Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'`caused by mixing `import` and `module.exports` syntax while code transpiling using `babel`.
The problem is described [here](https://github.com/webpack/webpack/issues/4039)

I tried manually replace `module.exports` to `export default` in `modulles/csv/src/bundle.js `file and it becomes work.
So As I understood we are having the options to fix this issue:
1) Use `umd` standard like this `['@babel/preset-env', {modules: 'umd'}]` - This is a common module format which will be supported as by `browser` as by `NodeJs`. - **The preferred fix!**
1) Replace all `module.exports`  with `export default` in all files in project.
2) Use `"sourceType": "unambiguous"` flag.  It is described [here](https://babeljs.io/docs/en/options#sourcetype)

dryabinin-actionengine: While testing in `examples/gallery` there are no errors on`2.3.9` version. Looks like not all modules are broken.
If I try to fetch another module for example `csv` loader in `examples/gallery/progress.html` on version `2.3.9` I got the same error: `dist.min.js:1 Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'`
If I apply my changes and will try to load script locally from `modules/csv/dist/dist.min.js` there are no errors and `loaders` object exists and has `CSVLoader` inside.

image


image

ibgreen:
@dryabinin-actionengine We build such scripts in several of our frameworks. It could be worth comparing with setup in deck.gl - since these scripts are designed to work together, they probably should be built in similar ways.

dryabinin-actionengine:
> @dryabinin-actionengine We build such scripts in several of our frameworks. It could be worth comparing with setup in deck.gl - since these scripts are designed to work together, they probably should be built in similar ways.

I tried to use the same Babel structure as in `deck.gl`. But it is not working too.
Looks like this error occurs because of internal webpack logic:
As I said before this error happens not in every module.
For example in `csv` module we failed on `module.exports = Papa;` line in `papaparse.js` just because in another files we do `import Papa from './libs/papaparse';` We can't mix `import` and `module.export` syntax.
The same error will be in  `las` loader. It also has `module.exports` syntax in `libs/laz-perf.js `file mixing with `import`. That occurs the same error like in `csv` module.

In `deck.gl` we don't have we the same `module.exports` mixing with `import` logic, so no errors there and `Babel` config works good.

So I guess we should properly handle `module.exports` while `Babel` transpilation.
To avoid such errors I suggest the following options:

1) `['@babel/preset-env', {modules: 'umd'}]` - [spec](https://babeljs.io/docs/en/babel-preset-env#modules)
2)  `["@babel/plugin-transform-modules-commonjs", {
      "allowTopLevelThis": true
    }]
]` - [spec](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs) 


All tests I do using `gallery/progress.html` file where I load modules using `<script>` tags.

ibgreen:
> In deck.gl we don't have we the same module.exports mixing with import logic, so no errors there and Babel config works good.

How about we fix the imports in the CSV and LAS modules? I would prefer that if possible.

ibgreen:
Good findings, by the way!

dryabinin-actionengine:
> > In deck.gl we don't have we the same module.exports mixing with import logic, so no errors there and Babel config works good.
> 
> How about we fix the imports in the CSV and LAS modules? I would prefer that if possible.

Done. Modules were fixed and tested one by one using `examples/gallery `app. Every script was loaded and checked global `loaders` object with loader inside.

image

Pessimistress (approved):

ibgreen:
Don't forget to cherry-pick to master

PR #1155, opened 2021-02-05
Change sharedResource href structure as per spec

dryabinin-actionengine:
`sharedResouce` should be an `object`
https://github.com/Esri/i3s-spec/blob/master/docs/1.7/3DNodeIndexDocument.cmn.md


ibgreen (approved):
I will always propagate for adding a unit test. The fact that we had to fix this shows the value.

PR #1181, opened 2021-02-15
gltf: when resolving accessors in post-processing, take into account accessor.byteOffset

dryabinin-actionengine:
There is conflict while cherry-pick process of [PR](https://github.com/visgl/loaders.gl/pull/1175) to 2.3-release branch.
@Avnerus @ibgreen Please Take a look at thit.

### DIFF
--- a/modules/gltf/src/lib/post-process-gltf.js
+++ b/modules/gltf/src/lib/post-process-gltf.js
@@ -288,7 +288,8 @@ class GLTFPostProcessor {
     if (accessor.bufferView) {
       const buffer = accessor.bufferView.buffer;
       const {ArrayType, length} = getAccessorArrayTypeAndLength(accessor, accessor.bufferView);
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
Not sure what the conflict was but this looks good to me.

dryabinin-actionengine:
@ibgreen @Avnerus If I cherry-pick commit I got the following conflict:
image
What` accessor.value `we should use?
1) `accessor.value = new ArrayType(buffer.arrayBuffer, byteOffset, length);`

Avnerus:
Hi,
I think (1) might be a bit better because `slice` returns a copy while the constructor method creates a view?

dryabinin-actionengine:
> Hi,
> I think (1) might be a bit better because `slice` returns a copy while the constructor method creates a view?

Ok, Thanks!

PR #1183, opened 2021-02-16
tiles: Tile traversal optimizations

Avnerus:
Hi! this is a long one.
As part of our work on Cesium 3d tiles with photogrammetry and three.js I ended up doing some changes to the loaders.gl tiles code which I detail in this PR:

1. Consolidate the tile `getPriority` function:
When working with heavy street-view-style photogrammetry tilesets, I always had to use `requestThrottling`, or the browser could not handle the amount of tiles being loaded. However, the priority function used for the request scheduler was using only `lodMetricValue`, not taking into account the distance to the camera. A priority function which resembles more the [CesiumJS function](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Cesium3DTile.js#L1660) and takes both sse and camera distance into account was used in the [traversal code](https://github.com/visgl/loaders.gl/blob/master/modules/tiles/src/tileset/traversers/tileset-traverser.js#L270l) when loading the tile, I therefore moved that code to the tile and use that one instead, both on tile load and for request scheduling.

2. Regarding Skip VS Base traversal, we do not yet have a skip traversal implementation in three.js, which requires the use of the stencil buffer for tile depth (do you by the way have an implementation in deck.gl or other?). However, I have done some tests and noticed that there might be some lines in the traversal and priority functions that are specific for either skip or base traversal: 
https://github.com/visgl/loaders.gl/blob/480b38a9a2cc9e3ed19c7ecb0aea9665add8901f/modules/tiles/src/tileset/traversers/tileset-traverser.js#L172-L174
This line breaks the iteration of a parent tile if at least one child is not yet refined. I think this is specific for base traversal because in skip traversal you might want to load a certain child even if siblings are not loaded. This break does not happen in the [original CesiumJS code](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Cesium3DTilesetTraversal.js#L486-L501). I then changed it to:
```        
refines = refines && childRefines;
if (!refines && !skipLevelOfDetail) {
  return false;
}
```
And then here:
https://github.com/visgl/loaders.gl/blob/480b38a9a2cc9e3ed19c7ecb0aea9665add8901f/modules/tiles/src/tileset/tile-3d.js#L160-L164
This line aborts the loading of a tile if it is not visible, but in base traversal you might need to load invisible tiles or else the parent would not refine. I had to change it to:
```
if (!this.isVisible && traverser.options.skipLevelOfDetail) {
  return -1;
}
```
So I could enable `requestThrottling`.

3. Finally, I added to safety checks related to tile unloading. This is related to the fact that if in order to get the traversal working properly in three.js, I had to intervene in the `contentState` process and decide myself when the tiles are definitely ready once all of the three.js objects are created (can be after an asynchronous gltf parsing operation). This intervening introduced some race conditions in which a tile's content was being accessed even after it was destroyed. Perhaps the proper way to fix this is to support a user defined `Promise` which defers the tile's `contentState` setting of `READY` or `UNLOADED`, but in the meantime I think it is enough to have those race conditions checks.

No urgency on this one as we are using our fork for now, but would appreciate your comments on these fixes.
Thank you!

/Avner

### DIFF
--- a/modules/tiles/docs/api-reference/tileset-3d.md
+++ b/modules/tiles/docs/api-reference/tileset-3d.md
@@ -69,7 +69,8 @@ Parameters:
 - `json`: loaded tileset json object, should follow the format [tiles format](https://loaders.gl/docs/specifications/category-3d-tiles)
 - `options`:
   - `options.ellipsoid`=`Ellipsoid.WGS84` (`Ellipsoid`) - The ellipsoid determining the size and shape of the globe.
-  - `options.throttleRequests`=`true` (`Boolean`) - Determines whether or not to throttle tile fetching requests.
+  - `options.throttleRequests`=`true` (`Boolean`) - Determines whether or not to throttle tile fetching requests. Throttled requests are prioritized according to tile visibility.
+  - `options.maxRequests`=`64` (`Number`) - When throttling tile fetching, the maximum number of simultaneous requests.
   - `options.modelMatrix`=`Matrix4.IDENTITY` (`Matrix4`) - A 4x4 transformation matrix this transforms the entire tileset.
   - `options.maximumMemoryUsage`=`512`] (`Number`) - The maximum amount of memory in MB that can be used by the tileset.
   - `options.fetchOptions` - fetchOptions, i.e. headers, used to load tiles from tiling server
--- a/examples/website/i3s/app.js
+++ b/examples/website/i3s/app.js
@@ -144,7 +144,7 @@ export default class App extends PureComponent {
 
   _renderLayers() {
     const {tilesetUrl, token} = this.state;
-    const loadOptions = {throttleRequests: true};
+    const loadOptions = {};
     if (token) {
       loadOptions.token = token;
     }
--- a/modules/tiles/docs/README.md
+++ b/modules/tiles/docs/README.md
@@ -1,4 +1,4 @@
-# Overview
+#s Overview
 
 `@loaders/tiles` exposes handy classes `Tileset3D` and `Tile3D` which can understand the loaded data from tile loaders (`@loaders.gl/3d-tiles`, `@loaders.gl/i3s`, etc.), and provide useful functions for dynamically selecting tiles for rendering under a viewport.
 
@@ -38,7 +38,7 @@
   - `options`: Options object, but not limited to
     Parameters:
     - `modelMatrix`=`Matrix4.IDENTITY` (`Matrix4`) - A 4x4 transformation matrix that transforms the tileset's root tile.
-    - `maximumMemoryUsage`=`512`] (`Number`) - The maximum amount of memory in MB that can be used by the tileset.
+    - `maximumMemoryUsage`=`512` (`Number`) - The maximum amount of memory in MB that can be used by the tileset.
     - `ellipsoid`=`Ellipsoid.WGS84` ([`Ellipsoid`](https://math.gl/modules/geospatial/docs/api-reference/ellipsoid)) - The ellipsoid determining the size and shape of the globe.
       Callbacks:
     - `onTileLoad` (`(tileHeader : Tile3DHeader) : void`) - callback when a tile node is fully loaded during the tileset traversal.
--- a/modules/tiles/docs/README.md
+++ b/modules/tiles/docs/README.md
@@ -1,4 +1,4 @@
-#s Overview
+# Overview
 
 `@loaders/tiles` exposes handy classes `Tileset3D` and `Tile3D` which can understand the loaded data from tile loaders (`@loaders.gl/3d-tiles`, `@loaders.gl/i3s`, etc.), and provide useful functions for dynamically selecting tiles for rendering under a viewport.
 
--- a/modules/tiles/docs/api-reference/tile-3d.md
+++ b/modules/tiles/docs/api-reference/tile-3d.md
@@ -42,9 +42,9 @@ One of
 - `render`: has content to render
 - `tileset`: tileset tile
 
-###### `depth` (Number)
+##### `_selectionDepth` (Number)
 
-The depth of the tile in the tileset tree.
+The depth of the tile in the traversal tree.
 
 ###### `content` (Object)
 
--- a/modules/tiles/src/tileset/tile-3d.ts
+++ b/modules/tiles/src/tileset/tile-3d.ts
@@ -50,7 +50,7 @@ export default class TileHeader {
   id: string;
   url: string;
   parent: TileHeader;
-  refine: string;
+  refine: number;
   type: string;
   contentUrl: string;
   lodMetricType: string;
@@ -86,7 +86,7 @@ export default class TileHeader {
   private _centerZDepth: number;
   private _screenSpaceError: number;
   private _visibilityPlaneMask: any;
-  private _visible: boolean;
+  private _visible?: boolean;
   private _inRequestVolume: boolean;
 
   private _stackLength: number;
@@ -178,7 +178,7 @@ export default class TileHeader {
     this._shouldRefine = false;
     this._distanceToCamera = 0;
     this._centerZDepth = 0;
-    this._visible = false;
+    this._visible = undefined;
     this._inRequestVolume = false;
     this._stackLength = 0;
     this._selectionDepth = 0;
@@ -289,6 +289,44 @@ export default class TileHeader {
     }
   }
 
+  /*
+   * If skipLevelOfDetail is off try to load child tiles as soon as possible so that their parent can refine sooner.
+   * Tiles are prioritized by screen space error.
+   */
+  _getPriority() {
+    const traverser = this.tileset._traverser;
+    const {skipLevelOfDetail} = traverser.options;
+
+    /*
+     * Tiles that are outside of the camera's frustum could be skipped if we are in 'ADD' mode
+     * or if we are using 'Skip Traversal' in 'REPLACE' mode.
+     * In 'REPLACE' and 'Base Traversal' mode, all child tiles have to be loaded and displayed,
+     * including ones outide of the camera frustum, so that we can hide the parent tile.
+     */
+    const maySkipTile = this.refine === TILE_REFINEMENT.ADD || skipLevelOfDetail;
+
+    // Check if any reason to abort
+    if (maySkipTile && !this.isVisible && this._visible !== undefined) {
+      return -1;
+    }
+    if (this.contentState === TILE_CONTENT_STATE.UNLOADED) {
+      return -1;
+    }
+
+    // Based on the priority function `getPriorityReverseScreenSpaceError` in CesiumJS. Scheduling priority is based on the parent's screen space error when possible.
+    const parent = this.parent;
+    const useParentScreenSpaceError =
+      parent && (!maySkipTile || this._screenSpaceError === 0.0 || parent.hasTilesetContent);
+    const screenSpaceError = useParentScreenSpaceError
+      ? parent._screenSpaceError
+      : this._screenSpaceError;
+
+    const rootScreenSpaceError = traverser.root ? traverser.root._screenSpaceError : 0.0;
+
+    // Map higher SSE to lower values (e.g. root tile is highest priority)
+    return Math.max(rootScreenSpaceError - screenSpaceError, 0);
+  }
+
   /**
    *  Requests the tile's content.
    * The request may not be made if the Request Scheduler can't prioritize it.
@@ -394,7 +432,6 @@ export default class TileHeader {
     this._visible = this._visibilityPlaneMask !== CullingVolume.MASK_OUTSIDE;
     this._inRequestVolume = this.insideViewerRequestVolume(frameState);
 
-    this._priority = this.lodMetricValue;
     this._frameNumber = frameState.frameNumber;
     this.viewportIds = viewportIds;
   }
@@ -594,7 +631,7 @@ export default class TileHeader {
     this._centerZDepth = 0;
     this._screenSpaceError = 0;
     this._visibilityPlaneMask = CullingVolume.MASK_INDETERMINATE;
-    this._visible = false;
+    this._visible = undefined;
     this._inRequestVolume = false;
 
     this._stackLength = 0;
@@ -607,18 +644,6 @@ export default class TileHeader {
     this._requestedFrame = 0;
 
     this._priority = 0.0;
-  }
-
-  _getPriority() {
-    // Check if any reason to abort
-    if (!this.isVisible) {
-      return -1;
-    }
-    if (this.contentState === TILE_CONTENT_STATE.UNLOADED) {
-      return -1;
-    }
-
-    return Math.max(1e7 - this._priority, 0) || 0;
   }
 
   _getRefine(refine) {
--- a/modules/tiles/src/tileset/tileset-3d.ts
+++ b/modules/tiles/src/tileset/tileset-3d.ts
@@ -65,6 +65,7 @@ export type Tileset3DProps = {
   token?: string;
   attributions?: string[];
   headers?: any;
+  maxRequests?: number;
   loadTiles?: boolean;
   fetchOptions?: {[key: string]: any};
   basePath?: string;
@@ -84,6 +85,7 @@ type Props = {
   token: string;
   attributions: string[];
   headers: any;
+  maxRequests: number;
   loadTiles: boolean;
   fetchOptions: {[key: string]: any};
   basePath: string;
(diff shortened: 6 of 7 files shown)


ibgreen:
@Avnerus Really excited to see that you are making such in-depth use of the 3D Tiles loader. 

Regarding your changes, they make perfect sense though I have to admit that they do require quite a bit of effort and context switching for me to review as it has been a while since I worked on this module, and we currently do not really have an active maintainer of the 3D tiles traversal parts. 

Since you have done enough research to make changes of this caliber and you are maintaining your own fork of loaders.gl, how would you feel about becoming a maintainer (or even lead) of the 3d tiles module in loaders.gl?

As a maintainer you'd get write access to loaders.gl repo, ability to approve PRs, influence direction etc. We also have public open governance meetings and also steering committee meetings etc that you could join at your option.

Note that we do have active full-time maintainers on the I3SLoader and also the rest of loaders.gl, it is mainly the 3D tiles area that is poorly covered at the moment, so your scope would be limited. 

If you are interested in learning more about how this could work, we can chat on the deck.gl forum or contact me via email in my profile.

Avnerus:
Hi @ibgreen, that sounds really interesting. I'll send you an e-mail.
Thank you!
/Avner

Avnerus:
Added to this PR fixes for empty traversal. The refinement check was not working correctly for nested `tileset.json`s. I left still some _TODOs_ to keep an eye on, but with this fix the nested traversal seems to be working well.

modules/tiles/src/tileset/traversers/tileset-traverser.js
```diff
@@ -145,6 +144,8 @@ export default class TilesetTraverser {
       !skipLevelOfDetail && tile.refine === TILE_REFINEMENT.REPLACE && tile.hasRenderContent;
 
     let hasVisibleChild = false;
+    let refines = true;
```
ibgreen, line 147: Should we make refinement calculation into a separate helper function? And ideally add a unit test case?

modules/tiles/src/tileset/traversers/tileset-traverser.js
```diff
@@ -307,28 +284,36 @@ export default class TilesetTraverser {
     let allDescendantsLoaded = true;
     const stack = this._emptyTraversalStack;
 
+    stack.push(root);
+
     while (stack.length > 0) {
       const tile = stack.pop();
 
+      /* TODO: This doesn't seem needed, should check further.
```
ibgreen, line 292: Would be great if we can sort this out.

ibgreen (approved):
This looks good. 
- One of the biggest problems with our traversal is lack of unit tests - which makes any refactor of the already complicated code tricky to review. Any test you could add, no matter how trivial, would help.
- We need to check that these changes work in I3S as well, as the `tiles` module is shared between both formats. You can start by running the I3S example, and @dryabinin-actionengine could also help you test.

modules/tiles/src/tileset/traversers/tileset-traverser.js
Avnerus: Hi, about a helper function, at least for me it helps to try and mirror the [CesiumJS code](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Cesium3DTilesetTraversal.js#L441-L529) as much as possible so I can somewhat keep track of the differences.
Unit testing would definitely be good, but I will not have the time to add that soon.

modules/tiles/src/tileset/traversers/tileset-traverser.js
Avnerus: So far, at least in my test cases, it seems to work well. The tiles are updated in the main traversal and I don't see any glitches.

Avnerus:
Regarding I3S, I can confirm at least that the [examples here](https://github.com/visgl/loaders.gl/tree/master/examples/website/i3s) seem to work. but would be good if someone could make a performance comparison before/after the fixes. Also you could try adding `throttleRequests: true` to the options, which I believe should work correctly now.

ibgreen:
@belom88 @dryabinin-actionengine Please test on I3S and land this PR asap.

modules/tiles/src/tileset/traversers/tileset-traverser.js
Avnerus: These turned out to be needed after-all in a more complex model.

belom88:
Looks like traverser requests much more tiles than before during refinement.
1. Switch to New York;
2. Wait for "tiles loading: 0";
3. Zoom in on particular area;
4. Tiles loading starts from 140-150.

https://drive.google.com/file/d/1HJWoSHTf8cViES0AJntC075cpXpObyEY/view?usp=sharing

Expected behavior (on master):
...
4. Tiles loading starts from 70-80 and loading time is less.

https://drive.google.com/file/d/1ECQEQ7bhzhAZgwhwbnFAMOng5SdXIg0M/view?usp=sharing

Avnerus:
Hi @belom88!
This is not yet the final PR for merge, but I updated the branch with my latest works and fixes on tiles.
Could you perhaps run another test on i3s and let me know how is the performance on your end? On my tests I seem to be getting equal or better results.
Thank you!
/Avner

modules/worker-utils/src/lib/worker-farm/worker-farm.d.ts
```diff
@@ -19,7 +19,6 @@ export default class WorkerFarm {
   /** Get the singleton instance of the global worker farm */
   static getWorkerFarm(props: WorkerFarmProps): WorkerFarm;
 
-  readonly maxConcurrency: number;
```
ibgreen, line 22: Why is this deleted? This should be specifiable on the farm?

modules/worker-utils/src/lib/worker-farm/worker-farm.js
```diff
@@ -41,12 +41,20 @@ export default class WorkerFarm {
 
   getWorkerPool({name, source, url}) {
     let workerPool = this.workerPools.get(name);
+
+    const maxConcurrency = isMobile ? this.props.maxMobileConcurrency : this.props.maxConcurrency;
+
+    if (workerPool && workerPool.maxConcurrency !== maxConcurrency) {
```
ibgreen, line 47: Why?

ibgreen (commented):
@Avnerus Thanks for getting back to this. I think it would be great if you could split this into something like 3-4 PRs:

- one simple PR bumping the dependencies and making any necessary changes to the webpack setup etc that we can land quickly (so you can rebase the remaining PRs)
- one PR with the request throttling changes
- one PR with the worker farm changes
- one PR with the more delicate traversal changes

Avnerus:
Hi @ibgreen, sorry I didn't clarify this. This branch now contains a copy of my fork, which will be split and cleaned. I just wanted to quickly get something out there so @belom88 is able to test the performance on i3s which I now see as the main issue blocking the changes, while I continue working on the PRs.

modules/worker-utils/src/lib/worker-farm/worker-farm.js
Avnerus: This is code that I still need to revisit, because it worked only before the worker logic was refactored. the goal here is to be able to change the `maxConcurrency` in runtime. I found it useful to be able to quickly test different concurrency configurations using the gui.

ibgreen: Perhaps we could make `WorkerFarm.setProps({maxConcurrency})` call workerPool.setProps({maxConcurrency}) on all  workerPools?

If so, WorkerPools may need some logic to destroy returned workers if the maxConcurrency has been reduced.

Avnerus: Yes this might be the missing piece. I will work on that one later on. Thanks!

belom88:
I checked it. Performance is OK now.
Another good thing is that it fixes overlapping tiles:
Before:
image (4)
After:
image

It was mentioned by Alexey Polyashov in Slack.

Avnerus:
Super! 
Thank you for the update @belom88.

ibgreen:
@Avnerus This is great news. Since we are wrapping up 3.0 it would be fantastic if we can get these changes landed soon. I'll prioritize review of any PRs you open.

Avnerus:
Thank you! l believe by the end of next week I should have the final PRs ready. 
My work also depends on #1436 so hopefully that would also be done in time for 3.0

ibgreen:
@Avnerus Just to be confirm, sounds like you are not planning to do a PR for  #1436 but would like us to do it.

Avnerus:
> @Avnerus Just to be confirm, sounds like you are not planning to do a PR for #1436 but would like us to do it.

I undertook from @dryabinin-actionengine that he is already working on kt

modules/worker-utils/src/lib/worker-farm/worker-farm.js
Avnerus: Hi @ibgreen!
I revisited this code now and it does actually work. Earlier I just forgot to alias the new worker-utils module.
The `getWorkerFarm` method already calls `setProps`, so eventually `maxConcurrency` propagates from the load options down to the worker pool. Essentially with this code, every time a loader is passed a `maxConcurrency` that is different from the existing value, the worker threads are destroyed and then recreated with the updated number.
Do you think this solution can work or do you foresee an issue?

ibgreen: I don't see it. 
- The workerFarm just passes the props when it creates a new workerPool, not during workerFarm.setProps. 
- I think it could work if workerFarm iterated over all its current workerPools in `setProps`.

Avnerus:
@ibgreen and @belom88, this PR now contains only the tile traversal changes. Here is recap and summary of the changes:

1. Consolidate the `getPriority` function to be used by both the traverser and request scheduler. Set `requestThrottling` as `true` by default, increase `maxRequests` to 64 by default and allow changing it via options.
2. Fix the empty traversal method which previously didn't work well (used when there are nested tileset jsons).
3. Minor changes in traversal so it matches the CesiumJS code and some extra safe-checks.
4. Revised documentation for `throttleRequests` and documentation for `maxRequests`.

Let me know your comments!

Thank yo

modules/tiles/docs/api-reference/tileset-3d.md
```diff
@@ -68,8 +68,9 @@ Parameters:
 
 - `json`: loaded tileset json object, should follow the format [tiles format](https://loaders.gl/docs/specifications/category-3d-tiles)
 - `options`:
-  - `options.ellipsoid`=`Ellipsoid.WGS84` (`Ellipsoid`) - The ellipsoid determining the size and shape of the globe.
-  - `options.throttleRequests`=`true` (`Boolean`) - Determines whether or not to throttle tile fetching requests.
+  - `options.ellipsoid`=`Ellipsoid.WGS84` Determines whether or not to throttle tile fetching requestDetermines whether or not to throttle tile fetching request(`Ellipsoid`) - The ellipsoid determining the size and shape of the globe.
```
ibgreen, line 71: This looks strange, copy paste?

modules/tiles/src/tileset/tile-3d.js
```diff
@@ -170,16 +170,32 @@ export default class TileHeader {
     }
   }
 
+  // If skipLevelOfDetail is off try to load child tiles as soon as possible so that their parent can refine sooner.
```
ibgreen, line 173: JSDoc style `/** */` shows the comment when hovering in vscode.

modules/tiles/src/tileset/tile-3d.js
```diff
@@ -461,7 +478,7 @@ export default class TileHeader {
     this._centerZDepth = 0;
     this._screenSpaceError = 0;
     this._visibilityPlaneMask = CullingVolume.MASK_INDETERMINATE;
-    this._visible = false;
+    this._visible = null;
```
ibgreen, line 481: What is the type of _visible? Can it be set to `null`? We have a typescript PR about to land.

modules/tiles/src/tileset/tileset-3d.js
```diff
@@ -551,7 +555,7 @@ export default class Tileset3D {
   }
 
   _unloadTile(tile) {
-    this.gpuMemoryUsageInBytes -= tile.content.byteLength || 0;
+    this.gpuMemoryUsageInBytes -= (tile.content && tile.content.byteLength) || 0;
```
ibgreen, line 558: It would be good to maintain a `gpuMemoryUsageInBytes` on the tile so we didn't need to rake into its internals here. I am not sure `content` would usually be a typed array.

modules/tiles/src/tileset/tile-3d.js
```diff
@@ -170,16 +170,32 @@ export default class TileHeader {
...
+    if (maySkipTile && !this.isVisible && this._visible !== null) {
       return -1;
     }
     if (this.contentState === TILE_CONTENT_STATE.UNLOADED) {
       return -1;
     }
 
-    return Math.max(1e7 - this._priority, 0) || 0;
+    const parent = this.parent;
+    const useParentScreenSpaceError =
```
ibgreen, line 189: If you are able to add some more comments (given that you have spent time understanding the algorithm) that would be nice. Imagine the questions a new person would have when looking at this code.

ibgreen (approved):
Looks pretty clean, given that I3S testing checks out I am fine with this.

modules/tiles/src/tileset/tile-3d.js
Avnerus: Would it work as optional `_visible?: boolean`? 
I used this to indicate when the visibility is not at all updated (headless, tile-covnerter) so the scheduler will still process the request.

modules/tiles/src/tileset/tileset-3d.js
Avnerus: I am not sure what you mean? All I added here was a check that `tile.content` exists (I experienced a race condition in which the content was already gone) .

Avnerus:
@ibgreen Added another change which I understood should be attached to this PR.
`tile._selectionDepth` indicates the traversal depth of the tile. This is mostly useful if trying to implement Skip Traversal with the use of the ZBuffer or other method for hiding tiles according to depthl
I added it to the documentation. There was a `depth` property in the documentation but it didn't really exist.
`_selectionDepth` is how it is called in CesiumJS.

modules/tiles/docs/api-reference/tileset-3d.md
Avnerus: Yes, fixed!

modules/tiles/src/tileset/tile-3d.js
Avnerus: Fixed!

modules/tiles/src/tileset/tile-3d.js
ibgreen: Well OK, I certainly wish we didn't need so many optional fields. The firmer we can type a field the better. More variations => more bugs.

modules/tiles/src/tileset/tileset-3d.js
ibgreen: Sorry I see that I wasn't reviewing your change, I was reviewing the original code.

modules/tiles/src/tileset/tile-3d.js
Avnerus: Well, another way would be to add a second boolean such as `headless` which would be falsified when visibility is updated, but I am not sure if it is more readable.
How about using `undefined` and defining in Typescript:
`_visible: boolean | undefined` ? I'm not sure what's the standard way to type variables that have an optional value (like this [Rust equivalent)](https://doc.rust-lang.org/std/option/) or [this one](https://gcanti.github.io/fp-ts/modules/Option.ts.html)).

modules/tiles/src/tileset/tileset-3d.js
Avnerus: Ah ok. we can the maybe handle this later on.

modules/tiles/src/tileset/tile-3d.js
ibgreen: Sure. If it is only this field that needs three states, then `_visible?: boolean` would work as opposed to having two fields just for this. However if we need to return this to the `undefined` state at some point, I'd rather use `null` (`_visible: boolean | null`. 

If there were multiple fields that could be controlled by one common field that indicates whether things are initialized, then I'd personally separate it out.

This is all moot before we actually land the typescript branch.

Avnerus: OK. I think then we can stick with `_visible?: boolean` for this one. Seems readable enough to me.

modules/tiles/src/tileset/tile-3d.js
Avnerus: Added some comments.

Avnerus:
@ibgreen  @belom88 , this is now rebased to Typescript.

belom88 (commented):
1. Cannot build the website on this branch;
2. If I try to do `git merge master`, there are conflicts

PR #1184, opened 2021-02-16
Comment out octet-stream MIMEtype for the MVTLoader

dryabinin-actionengine:


kylebarron (approved):
I personally think it would be cleaner to remove it altogether for now instead of comment it, but either way is ok

dryabinin-actionengine:
It was published as 2.3.12 version.

PR #1234, opened 2021-03-02
I3S picking - Select building using only onclick event

dryabinin-actionengine:
- Enabled picking color only when user click on object (building).
- After this changes we get an issue in `I3S v1.7` picking functionality:

In` I3s v1.6 `vesrion we are using `featureIds` which were created by using `id` and `faceRange` data of particular tile.
This `featureIds` represents `OBJECTID` values from `attributeStorageInfo` and they are all unique across all tiles.
So when we encode such `featureIds` to `pickingColors` we have unique color for each building across all tiles on the map.

In `I3S v1.7` version `featureIds` from compressed geometry are like feature indices for every attribute in `attributeStorageInfo`. We can find particular building attributes using `featureId` as index in `attributeStorageInfo` object. So such `featureIds` are not unique across all tiles in tileset. It only unique across only one tile. When we encode such `featureIds` as indices to `pickingColors` we don't have a unique color for each building across all tiles on the map.

The problem is when we want to select only one building on the map in `i3s v1.7`  we can see that all objects which has the same `featureId` are selected.

@ibgreen Upd: [pull request](https://github.com/visgl/loaders.gl/pull/1236) to fix that ^^^

Screenshot from 2021-03-02 15-56-32

### DIFF
--- a/examples/experimental/i3s-picking/app.js
+++ b/examples/experimental/i3s-picking/app.js
@@ -77,7 +77,8 @@ export default class App extends PureComponent {
       name: INITIAL_EXAMPLE_NAME,
       viewState: INITIAL_VIEW_STATE,
       selectedMapStyle: INITIAL_MAP_STYLE,
-      selectedFeatureAttributes: null
+      selectedFeatureAttributes: null,
+      selectedFeatureIndex: -1
     };
     this._onSelectTileset = this._onSelectTileset.bind(this);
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
I see. It is a big pity to loose the highlight on hover. 

We kind of need two highlights - one for the hover and one for the click.

We could modify the shader to support two picking colors, or change the color in the geometry when we select.

PR #1249, opened 2021-03-05
i3s: debug application

belom88:
I made I3S debug application.
Now "Frustum viewport" is done there.
Known issues:
- The frustum is shown on both viewports. On the main viewport it looks like pink border. So the tileset doesn't refine on the main viewport when the user zooms in.
- Tiles refinement relies on the minimap viewport (9 zoom, bearing 0, pitch 0).
Screenshot from 2021-03-05 12-16-24


ibgreen (approved):
Is this based on #636? Should we close that PR?

Let us add this example to the website for now so that we can easily demo it.

I see a lot of tiles outside of the view frustum in the overhead view. This indicates either
- The traversal algorithm loads too much
- We have not yet limited the traversal to only the view frustum of the primary view.

If we are going to render the view frustum in the primary view (not sure that is helpful), I recommend adding
```
   parameters: {
      depthTest: false
   }
```

to that LineLayer (and rendering it last) so it renders on top of everything.

belom88:
I've done it based on https://github.com/visgl/deck.gl/tree/master/test/apps/frustum-cull . But this PR is quite similar.
You see a lot of tiles because the traversal algorithm relies on minimap viewport (zoom 9, bearing 0, pitch 0). So it loads all the map on the root level and never refine tiles when you zoom in.

belom88:
The frustum is loaded on the primary view. This is the first issue I described in the topic. It looks like pink border

PR #1258, opened 2021-03-09
I3S-picking - Add picking checkbox

dryabinin-actionengine:
We can manually enable/disable objects picking.
image


ibgreen (approved):
If we are not picking can we have autohighlight on hover?

PR #1260, opened 2021-03-09
I3s-picking - Rename internal function

dryabinin-actionengine:
`getAttributesFromTileByFeatureId -> getTileAttributesFromFeatureId`


ibgreen (approved):
Any other functions that could follow the same naming convention?

Do we have typescript `d.ts` files for the exported functions?

dryabinin-actionengine:
> Any other functions that could follow the same naming convention?
> 
> Do we have typescript `d.ts` files for the exported functions?

As for functions. I guess we can do some refactoring later.
`d.ts` filed will be added in further PR

PR #1284, opened 2021-03-19
I3S Converter - Fix params order in normalizeAttributesByIndicesRange function

dryabinin-actionengine:


ibgreen (approved):
Types and/or tests would have caught this. Just saying...

PR #1285, opened 2021-03-21
i3s: content worker

belom88:
It perceptibly increase I3S loading speed:

https://user-images.githubusercontent.com/18549629/111904189-9c5b9400-8a56-11eb-9548-701a344e3c02.mp4


ibgreen (approved):
Nice first step.

A big issue with worker based parsing is that any non-binary data transferred between worker and mainthread is serialized to string and back, defeating a lot of the gains from moving the processing to a worker thread in the first place.

First, you will want to check how big the options object you pass in is. You can call JSON.stringify(options).length just before calling the loader. If it is more than a few KB we should see if you really need all that data.

Also, uncompressing the binary strings in the worker as you do here will likely make that part slower, as all the strings will be TextEncoded and TextDecoded again to move them back to the main thread. Let's think about whether we can do such processing at a higher level.

PR #1294, opened 2021-03-29
I3SLoader - Add benchmark

dryabinin-actionengine:
image


ibgreen (approved):
If the loader only loads one file, then using `parse` instead of load will measure just parsing, rather than the fetch time which varies on test setup.

When optimizing we can't really affect the fetch time, so values will be more meaningful if we only bench the actual CPU compute.

PR #1295, opened 2021-03-29
I3S-Debug - Add semantic validator ui

dryabinin-actionengine:
image


ibgreen (approved):
Nice to be able to validate in the UI. Are we also adding batch validation (perhaps an option to the `tile-converter`?)

dryabinin-actionengine:
> Nice to be able to validate in the UI. Are we also adding batch validation (perhaps an option to the `tile-converter`?)

What do you mean by batch validation? Sure we can implement additional validations to tile-converter, but this changes is applicable onnly for i3s-debug app.

PR #1318, opened 2021-04-12
WIP: Move constants from i3s-17-and-debug to the tile-converter module

dryabinin-actionengine:


ibgreen (approved):
I am a bit uncertain about this change.

the tile-converter module was created as a node.js module and I think you saw the kind of issues we can get when importing it into browser code. But if you solved the issue I suppose it is OK for now.

But I suppose the question is - where does the rest of the common validation code go. If we move the constants, we should probably also move some of the validation functions?

PR #1329, opened 2021-04-16
Tile-Converter - Add "--validate-bounding-volumes" option to script

dryabinin-actionengine:


ibgreen (commented):
Nice. Don't forget to update docs, also some test - at least making sure conversion doesn't crash when activating the new option...

Q: Is it possible to run this without actually converting?

Maybe simply offer two options
```
--convert
--validate
```

And validation and conversion can be activated independently, and we can add more validation features to the `--validate` flag as we go.

ibgreen (approved):

dryabinin-actionengine:
> Nice. Don't forget to update docs, also some test - at least making sure conversion doesn't crash when activating the new option...
> 
> Q: Is it possible to run this without actually converting?
> 
> Maybe simply offer two options
> 
> ```
> --convert
> --validate
> ```
> 
> And validation and conversion can be activated independently, and we can add more validation features to the `--validate` flag as we go.

Yes I thought about different modes like:
- convert
- validate
It is not quick change unfortunately but we will do it later and also we should add more validations here. 
Good suggestions. Thanks.

PR #1334, opened 2021-04-20
fix babel-runtime version

belom88:
It affects only the local debug process. `yarn.lock` in website has correct version of @babel/runtime


ibgreen (approved):
You can also pin babel in package.json.

belom88:
I tried. It doesn't work. There are two versions of `@babel/runtime` is installed in that case. The correct version is older than breaking one.

PR #1337, opened 2021-04-22
I3S-attribute-loader - Fix for attributes loading

dryabinin-actionengine:
This fix allows to load all attributes and doesn't fail all loaded attributes if any of them are not fetched or parsed properly

### DIFF
--- a/modules/i3s/src/helpers/promises-handler.js
+++ b/modules/i3s/src/helpers/promises-handler.js
@@ -1,0 +1,22 @@
+export const REJECTED_STATUS = 'rejected';
+export const FULFILLED_STATUS = 'fulfilled';
+
+/**
+ * Handle list of promises based on browser compatibility
+ * @param {Promise[]} promises
+ * @returns {Promise}
+ */
+export function handlePromises(promises) {
+  return Promise.allSettled ? Promise.allSettled(promises) : Promise.all(promises.map(reflect));
+}
+
+function reflect(promise) {
+  return promise.then(
+    value => {
+      return {status: FULFILLED_STATUS, value};
+    },
+    error => {
+      return {status: REJECTED_STATUS, reason: error};
+    }
+  );
+}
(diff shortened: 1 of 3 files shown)


modules/i3s/src/helpers/promises-handler.js
```diff
@@ -0,0 +1,22 @@
+export const REJECTED_STATUS = 'rejected';
+export const FULFILLED_STATUS = 'fulfilled';
+
+/**
+ * Handle list of promises based on browser compatibility
+ * @param {Promise[]} promises
+ * @returns {Promise}
+ */
+export function handlePromises(promises) {
+  return Promise.allSettled ? Promise.allSettled(promises) : Promise.all(promises.map(reflect));
```
ibgreen, line 10: Which browser is having issues with allSettled? Maybe mention in comment?

We could also just polyfill `allSettled`?

modules/i3s/src/i3s-attribute-loader.js
```diff
@@ -55,7 +55,7 @@ export async function loadFeatureAttributes(tile, featureId, options = {}) {
     attributeLoadPromises.push(promise);
   }
   try {
-    attributes = await Promise.all(attributeLoadPromises);
+    attributes = await handlePromises(attributeLoadPromises);
```
ibgreen, line 58: Maybe add a comment explaining why we need to support failed promises?

ibgreen (approved):

modules/i3s/src/i3s-attribute-loader.js
dryabinin-actionengine: Ok. Actually some of attributes can be encoded badly or some attributes can't to be loaded. We should get as many attributes as possible to show available attributes for user.

modules/i3s/src/helpers/promises-handler.js
dryabinin-actionengine: `Promise.allSettled() `was implemented in `ECMAScript 2020, the 11th edition`. It has wide browser support but pretty much for newest versions of browsers:

image

For example `Promise.all() ` support table:
image

Yes I think we can add a polyfill for it.

PR #1357, opened 2021-04-30
I3S-Debug - Refactoring for tiles color

dryabinin-actionengine:
Changes to move coloring logic from `tile-layer.js` file to `app-debug.js `to prepare merging changes back to deck.gl.

### DIFF
--- a/examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
+++ b/examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
@@ -54,16 +54,7 @@ export default class TileLayer extends Tile3DLayer {
   _makeSimpleMeshLayer(tileHeader, oldLayer) {
     const content = tileHeader.content;
     const {attributes, modelMatrix, cartographicOrigin, texture, material} = content;
-    const {
-      pickable,
-      autoHighlight,
-      tileColorMode,
-      pickFeatures,
-      colorMap,
-      selectedTileId,
-      coloredTilesMap,
-      wireframe
-    } = this.props;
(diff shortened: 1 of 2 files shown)


examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
```diff
@@ -54,16 +54,7 @@ export default class TileLayer extends Tile3DLayer {
...
-      pickable,
-      autoHighlight,
-      tileColorMode,
-      pickFeatures,
-      colorMap,
-      selectedTileId,
-      coloredTilesMap,
-      wireframe
-    } = this.props;
+    const {pickable, autoHighlight, pickFeatures, wireframe, getSimpleMeshLayerColor} = this.props;
```
ibgreen, line 57: `getSimpleMeshLayerColor` is a weird name.  I feel that the fact we are rendering a SimpleMeshLayer is an implementation detail. I would choose a name that is more semantic. `getMeshColor`, `getTileColor`, `getColor` etc.

ibgreen (approved):

examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
dryabinin-actionengine: Yes I supposed that you asked me about that. Before naming selection I looked on naming in Tile3DLayer in deck.gl.
For example for pointCloundLayer it uses `getPointColor` accessor:
image
I followed the same naming logic for simpleMeshLayer:
image

I suppose `getMesh` color is more accurate here.

PR #1375, opened 2021-05-11
I3S-Debug - Fix for i3s-content-loader worker

dryabinin-actionengine:
1. Move `segmentationData` to tile.content because we can't mutate tile object if we use` i3s-content-loader` worker.
2. Eliminate duplicate creation of object properties.


ibgreen (approved):
I am getting a bit confused, so let's clear things up. How does the final loaded structure look?

We have
- a tileset that has a field that contains an tree of tile tileHeaders?
- each tileHeader has a `content` field that points to the output of the tile loader
- are we adding a `tileset.<tileHeader>.content.content` field here?

If so, that would seem confusing. Clarifying the structure if the loaded data in the docs, with typescript etc would seem to be a relatively high priority.

dryabinin-actionengine:
> I am getting a bit confused, so let's clear things up. How does the final loaded structure look?
> 
> We have
> 
> * a tileset that has a field that contains an tree of tile tileHeaders?
> * each tileHeader has a `content` field that points to the output of the tile loader
> * are we adding a `tileset.<tileHeader>.content.content` field here?
> 
> If so, that would seem confusing. Clarifying the structure if the loaded data in the docs, with typescript etc would seem to be a relatively high priority.

- Yes tileset has three of headers
- Yes each tileHeader has a content inside
- No I don't create one more content field I just put `segmentationData` to other content fields:
test

PR #1376, opened 2021-05-11
tiles: loadTiles option

belom88:
This is alternative for https://github.com/visgl/deck.gl/pull/5755 . If it's okay we can close the PR in deck.gl.

### DIFF
--- a/examples/experimental/i3s-17-and-debug/app-debug.js
+++ b/examples/experimental/i3s-17-and-debug/app-debug.js
@@ -176,9 +176,10 @@ export default class App extends PureComponent {
       tileInfo: null,
       selectedTileId: null,
       coloredTilesMap: {},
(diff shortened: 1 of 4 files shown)


belom88:
> This looks too hacky. You are making this property into an object rather than a boolean so that you can keep a reference and modify it externally.
> 
> I think we should have a `setProps` type system instead.
> 
> ```
> tileset3D = new Tileset3D({loadTiles: true}); // the default
> tileset3D.setProps({loadTiles: false});
> ```

Done

ibgreen (approved):
Tileset3D is a documented class. Changes to the API should be reflected in the docs.

belom88:
> Tileset3D is a documented class. Changes to the API should be reflected in the docs.

I'll update it in another PR

PR #1382, opened 2021-05-14
Add support for legacy B3DM file format

dryabinin-actionengine:
@ibgreen Please take a look at my idea to validate 28 bytes header to detect legacy or wrong `b3dm` file structure.
To be honest it not really sophisticated method to check that but I didn't find any code which was done for such kind of [test](https://github.com/visgl/loaders.gl/blob/master/modules/3d-tiles/test/lib/parsers/batched-model-3d-tile.spec.js#L44). Please let me know if where was such kind of validation earlier.

### DIFF
--- a/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-tile-tables.js
+++ b/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-tile-tables.js
@@ -8,20 +8,55 @@ const SIZEOF_UINT32 = 4;
 // eslint-disable-next-line max-statements
 export function parse3DTileTablesHeaderSync(tile, arrayBuffer, byteOffset) {
   const view = new DataView(arrayBuffer);
+  let batchLength;
 
   tile.header = tile.header || {};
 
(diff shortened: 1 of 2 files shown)


ibgreen:
Also check this code in CesiumJS - If we can add this we might be able to support older versions:

https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Batched3DModel3DTileContent.js#L276

dryabinin-actionengine:
> Also check this code in CesiumJS - If we can add this we might be able to support older versions:
> 
> https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Batched3DModel3DTileContent.js#L276

Yes it works. I got a first case of legacy `b3dm` file format. After conversion I see New York. Geometry looks good but some buildings and wierd figures are flying in the air. I think it is made on purpose. Also attributes are broken.
image

Next items:
1. Add tests for two cases of legacy b3dm file format. 
2. Fix attributes if `b3dm` has legacy format.

ibgreen (approved):

PR #1385, opened 2021-05-17
i3s-debug: fix multiple viewports

belom88:
- Switch i3s-content-loader worker on;
- Fixes: `TypeError: t.getBoundingSphere(...).radius is not a function`


dryabinin-actionengine (approved):

kylebarron:
@belom88 looks like CI on `master` is failing because of this commit. Is that something you can debug?

belom88:
Yes. I switched worker on and tests with this worker can't load files from local path. https://github.com/visgl/loaders.gl/pull/1386 fixes tests

PR #1391, opened 2021-05-19
Extract earcut

felixpalmer:
visgl/loaders.gl#1356 implemented polygon triangulation using earcut in loaders.gl however seems more natural to have earcut in math.gl, so as to allow other projects to import it if needed. This PR moves earcut into math.gl, see https://github.com/uber-web/math.gl/pull/219 for more details


felixpalmer:
[WIP] Leaving this as draft until we land the change in math.gl (https://github.com/uber-web/math.gl/pull/219) so the dependency can be updated

kylebarron (approved):

alasarr (approved):

ibgreen (approved):

PR #1396, opened 2021-05-20
i3s: optimization

belom88:
This is improvement for I3S 1.7 loading:
- removed duplicated texture;
- draw indices instead of flattening arrays on CPU;
- optimization for creating MBS from OBB.

It works faster now. But we need to new release after merging this PR to have new version of worker.

### DIFF
--- a/modules/i3s/src/lib/parsers/parse-i3s-tile-content.js
+++ b/modules/i3s/src/lib/parsers/parse-i3s-tile-content.js
@@ -52,6 +52,9 @@ export async function parseI3STileContent(arrayBuffer, tile, tileset, options) {
   }
 
   tile.content.material = makePbrMaterial(tile.materialDefinition, tile.content.texture);
+  if (tile.content.material) {
+    tile.content.texture = null;
+  }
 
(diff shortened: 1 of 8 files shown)


modules/i3s/src/lib/parsers/parse-i3s-tile-content.js
```diff
@@ -373,15 +358,24 @@ function offsetsToCartesians(vertices, metadata = {}, cartographicOrigin) {
  * @returns {object}
  */
 function makePbrMaterial(materialDefinition, texture) {
-  if (!materialDefinition) {
-    return null;
+  let pbrMaterial;
```
Tamrat-B, line 361: +1 on creating default material when not present.

Tamrat-B (approved):
looks good.

ibgreen (approved):

ibgreen (commented):
Should we add some documentation to the loader explaining that materials will be created etc?

belom88:
> Should we add some documentation to the loader explaining that materials will be created etc?

Sure, I'll do it

PR #1428, opened 2021-06-02
geopackage: arrow schema

belom88:


modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -115,11 +147,12 @@ function getVectorTable(db, tableName, projections, {reproject, _targetCrs}) {
...
   }
 
+  const schema = getArrowSchema(db, tableName);
   if (projection) {
-    return transformGeoJsonCoords(geojsonFeatures, projection.project);
+    return {geojsonFeatures: transformGeoJsonCoords(geojsonFeatures, projection.project), schema};
   }
 
-  return geojsonFeatures;
+  return {geojsonFeatures, schema};
```
kylebarron, line 155: Note this is a breaking change because the final output from `parseGeoPackage` will change from
```js
{
	'table1': [features]
}
```
to
```js
{
	'table1': {
		'schema': schema,
		'geojsonFeatures': [features]
	},
	'table2' ...
}
```

It might be better if we standardized a geospatial output format for loaders to include both features and metadata. Might be good to discuss before the 3.0.0 release @ibgreen

ibgreen: Good point, let me look into that.

modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -381,3 +414,54 @@ function interleaveResults(columns, values) {
...
+    fields.push(new Field(columnMatch[1], new SQL_TYPES[columnMatch[2]](), true));
+    columnMatch = COLUMNS_REGEXP.exec(sql);
+  }
+  return new Schema(fields);
+}
+
+/**
+ * Form RegExp for taking column names and column types from SQL CREATE TABLE request
+ *
+ * @return {RegExp} an array of objects, where each object represents a single row
```
ibgreen, line 447: The description of this parameter looks wrong. This describes the action of the returned regexp rather than the return value.

ibgreen (approved):

modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -19,6 +31,26 @@ const ENVELOPE_BYTE_LENGTHS = {
...
+  INT: Int32,
+  INTEGER: Int32,
+  FLOAT: Float32,
+  DOUBLE: Float64,
+  REAL: Float64,
+  TEXT: Utf8,
+  BLOB: Binary,
+  DATE: Utf8,
+  DATETIME: Utf8,
+  GEOMETRY: Binary
```
kylebarron, line 49: While it's true that a geometry is internally stored as binary (WKB), that's an incorrect output schema type, since this loader converts it to GeoJSON.

modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -381,3 +414,54 @@ function interleaveResults(columns, values) {
...
+ * Get arrow schema
+ *
+ * @param  {Database} db GeoPackage object
+ * @param  {string} tableName  table name
+ * @return {Schema | null} Arrow-like Schema
+ */
+function getArrowSchema(db, tableName) {
+  const metadata = db.exec(
+    `SELECT * FROM \`sqlite_master\` WHERE type='table' AND tbl_name='${tableName}';`
+  )[0];
```
kylebarron, line 428: This query will be called on every table that exists in the GeoPackage. I haven't benchmarked, but I assume there's a non-negligible constant overhead to each SQL query.

Instead, I'd suggest doing a single query to retrieve the entire `sqlite_master` table (presumably it isn't much data) and then passing that result into each `getVectorTable` call. This would be similar to what I do for the projections table. I load the full set of projections _once_ and then pass that as an object into each `getVectorTable` call

https://github.com/visgl/loaders.gl/blob/97edd7ae1b236e8962c15dcfb5cedcfd1206aa07/modules/geopackage/src/lib/parse-geopackage.js#L60

https://github.com/visgl/loaders.gl/blob/97edd7ae1b236e8962c15dcfb5cedcfd1206aa07/modules/geopackage/src/lib/parse-geopackage.js#L66

modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -19,6 +31,26 @@ const ENVELOPE_BYTE_LENGTHS = {
...
+  MEDIUMINT: Int32,
+  INT: Int32,
+  INTEGER: Int32,
+  FLOAT: Float32,
+  DOUBLE: Float64,
+  REAL: Float64,
+  TEXT: Utf8,
+  BLOB: Binary,
+  DATE: Utf8,
+  DATETIME: Utf8,
```
kylebarron, line 48: Should we be decoding these `ISO-8601` dates into JS date objects?

modules/geopackage/src/lib/parse-geopackage.js
```diff
@@ -381,3 +414,54 @@ function interleaveResults(columns, values) {
...
+function getArrowSchema(db, tableName) {
+  const metadata = db.exec(
+    `SELECT * FROM \`sqlite_master\` WHERE type='table' AND tbl_name='${tableName}';`
+  )[0];
+  const sql =
+    (metadata.values &&
+      metadata.values[0] &&
+      metadata.values[0][4] &&
+      metadata.values[0][4].toString()) ||
+    '';
```
kylebarron, line 434: Might be worth a comment to say what `metadata.values[0][4]` is. The name `sql` doesn't mean much to me.

modules/geopackage/src/lib/parse-geopackage.js
kylebarron: Also, are we sure this is the simplest way to get type information about each column? I see that when you query `sqlite_master` you get the SQL string that would recreate the given table.

Why not use
```sql
SELECT name, type FROM pragma_table_info('<table_name>')
```

Then as output you get a simple mapping from column name to SQL type. No complex regex required.

```
+------------+----------+
| name       | type     |
+------------+----------+
| id         | INTEGER  |
| geom       | GEOMETRY |
| property_0 | TEXT     |
| property_1 | TEXT     |
| property_2 | TEXT     |
+------------+----------+
```

belom88: Done

belom88:
It is a question what we should generate schema for: input data or parsed data?
This is the result for now:
image

modules/geopackage/src/lib/parse-geopackage.js
belom88: Done

ibgreen (approved):
> It is a question what we should generate schema for: input data or parsed data?

@kylebarron Can you help resolve this question. As soon as that is sorted I think we can land this and create a task with follow up issues.

kylebarron:
> It is a question what we should generate schema for: input data or parsed data?
> This is the result for now:
> image

I would assume we want to generate the schema for the output data? This schema will be used by applications to understand how to interpret the data the application receives. So the application won't care about how the data was originally stored, just in what it receives I think. I'd guess that in most situations, however, the format of the input data will match the format of the output data.

I'm also reminded that we should (at some point, can be a future PR) handle mapping the column names from the _physical_ names to the _logical_ names. The GeoPackage spec allows for a separate name mapping, presumably so that if a user wants to rename a column, an `ALTER TABLE` isn't required, just a transaction in the table that holds name mappings. That's what this `dataColumns` field is, a mapping from physical name to logical name.

https://github.com/visgl/loaders.gl/blob/446643d6a8d19232a02baf85261769ce97119e9f/modules/geopackage/src/lib/parse-geopackage.js#L119

Then inside `constructGeoJsonFeature`, that mapping is applied on the columns

https://github.com/visgl/loaders.gl/blob/446643d6a8d19232a02baf85261769ce97119e9f/modules/geopackage/src/lib/parse-geopackage.js#L198-L202

kylebarron (approved):

ibgreen:
Open discussion items covered in #1443

PR #1435, opened 2021-06-03
Bounding Volume Layer - Refactoring

dryabinin-actionengine:
1. Save already generated meshes to the `tile.userData`
2. Code cleanup.


examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
```diff
@@ -3,7 +3,7 @@ import {Ellipsoid} from '@math.gl/geospatial';
 import {CubeGeometry, SphereGeometry} from '@luma.gl/engine';
 import {CompositeLayer, COORDINATE_SYSTEM, log} from '@deck.gl/core';
 import MeshLayer from './mesh-layer/mesh-layer';
-import {ORIENTED_BOUNDING_BOX, BOUNDING_SPHERE} from './constants';
+import {BOUNDING_VOLUME_MESH_TYPE} from './constants';
```
ibgreen, line 6: Is just `BOUNDING_VOLUME_TYPE` enough?

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
```diff
@@ -78,49 +85,37 @@ export default class BoundingVolumeLayer extends CompositeLayer {
       ...attributes.POSITION,
       value: new Float32Array(attributes.POSITION.value)
     };
+
```
ibgreen, line 88: Add some comments explaining what you are doing? If you are just translating the sphere, isn't that better done with `modelMatrix`?

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
```diff
@@ -78,49 +85,37 @@ export default class BoundingVolumeLayer extends CompositeLayer {
...
     const {content, viewportIds} = tile;
     const {material, getBoundingVolumeColor, boundingVolumeType} = this.props;
     const {cartographicOrigin, modelMatrix} = content;
+    const meshType = BOUNDING_VOLUME_MESH_TYPE[boundingVolumeType];
 
-    const geometry = this._getBoundingVolumeGeometry(tile, oldLayer, boundingVolumeType);
+    const mesh = tile.userData[meshType] || this._generateMesh(tile, meshType);
 
     return new MeshLayer({
       id: `obb-debug-${tile.id}`,
```
ibgreen, line 117:
```suggestion
      id: `bounding-volume-${tile.id}`,
```

examples/experimental/i3s-17-and-debug/constants.js
```diff
@@ -45,4 +45,9 @@ export const BOUNDING_VOLUME_TYPE = {
   OBB: ORIENTED_BOUNDING_BOX
 };
 
+export const BOUNDING_VOLUME_MESH_TYPE = {
```
ibgreen, line 48: Do we need a separate type for this?

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
```diff
@@ -78,49 +85,37 @@ export default class BoundingVolumeLayer extends CompositeLayer {
...
+    const sphereMesh = tile.header.mbs ? this._generateSphereMesh(tile) : null;
 
-    if (!mesh || mesh.constructor.name !== boundingVolumeType) {
-      return this._generateMesh(tile, boundingVolumeType);
-    }
+    tile.userData.cubeMesh = cubeMesh;
+    tile.userData.sphereMesh = sphereMesh;
 
-    return mesh;
+    return tile.userData[meshType];
```
ibgreen, line 105: Maybe put the meshes in a sub-object so that you can type it properly when we introduce typescript.

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
```diff
@@ -32,6 +31,16 @@ export default class BoundingVolumeLayer extends CompositeLayer {
     };
   }
 
+  updateState({changeFlags}) {
+    if (changeFlags.propsChanged) {
+      const {layerMap} = this.state;
+
+      for (const key in layerMap) {
```
ibgreen, line 38: This will update all layers on any prop change. Safe but perhaps some unnecessary work. Just watch out for perf issues.

ibgreen (approved):

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: `BOUNDING_VOLUME_TYPE` is already in use. This is the opposite map for `cubeMesh` and `sphereMesh`.
image

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: Yes. I tried to do update only if `boundingVolumeType` is changed. And it works better but we also provide a callback which is responsible for `boundingVolume` coloring. And if we change coloring mode in the debug app bounding volumes colors don't change. I didn't find a way to handle this. Maybe we can provide a coloring mode prop to the `boundingVolume` layer. If you remember in the debug app we have (random by tile, random by depth, user custom choice). But this prop will be useless for bounding-volume-layer just for checking is the `updateState` function.

examples/experimental/i3s-17-and-debug/constants.js
dryabinin-actionengine: I fell that existing bounding volume types are not good as object keys for tile.userData object.
```
export const BOUNDING_SPHERE = 'Bounding Sphere';
export const ORIENTED_BOUNDING_BOX = 'Oriented Bounding Box';
```

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: Do you mean mesh specific object inside` tile.userData`?

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
ibgreen: OK. It seemed to me that you just need an enum that distinguishes the types, and it is the same types in both cases. But this works too.

examples/experimental/i3s-17-and-debug/constants.js
ibgreen: I see. You could also make the types more general, and have a separate map for type names

```
enum BOUNDING_VOLUME_TYPE {
  BOUNDING_SHPERE,
  ORIENTED_BOUNDING_BOX
}

const BOUNDING_VOLUME_NAME = {
  [BOUNDING_VOLUME_TYPE.BOUNDING_SPHERE]: 'Bounding Sphere',
...
}
```

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
ibgreen: Yeah something like `tile.userData.boundingMeshes = {}`

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
ibgreen: OK.

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: We are tried to play with model matrix while creating mbs geometry and it doesn't work. Now we just pass `modelMatrix` as prop to the `MeshLayer` with generated mbs geometry. At this line we just do deep copy of `SphereGeometry` mesh to override it when we add center to each vertex.
Do you have better solution for this?

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: renamed

examples/experimental/i3s-17-and-debug/constants.js
dryabinin-actionengine: Will do in further PR

examples/experimental/i3s-17-and-debug/bounding-volume-layer.js
dryabinin-actionengine: Done

PR #1463, opened 2021-06-07
Fix for MVT benchmark

dryabinin-actionengine:


ibgreen (approved):
Just making sure that you know that you can test these with
- `yarn test bench`
- `yarn test bench-browser`

PR #1464, opened 2021-06-07
Use parseSync from loader context, not `@loaders.gl/core`

kylebarron:


kylebarron:
Hmm I guess `parseSync` isn't exported through the loader context?

ibgreen (approved):
Thanks! Don't forget to update the package.json...

kylebarron:
> Don't forget to update the package.json...

You already took that out in the last PR... https://github.com/visgl/loaders.gl/commit/678b464718e3228b89d36441a1133d37f0904742#diff-3efeb4ced94a131f3128972f9f7cb2f0e84e8d0515c8bcda886f249f75140699L27

https://github.com/visgl/loaders.gl/blob/0b3ed860fa863ce2f3b864b668d11cd2bf18ef2b/modules/geopackage/package.json#L26-L34

kylebarron:
It appears that currently the loader context only provides `fetch`, `url`, `parse`, and `loaders`, which seems to imply there's no way to do a `parseSync`. Inside the GeoPackage loader, `parseSync` with the `WKBLoader` is called on every small feature, so I’d guess avoiding the async overhead might be helpful here. Otherwise I can refactor to make parsing the internal WKB async.

PR #1466, opened 2021-06-08
Fix webpack configs for i3s example apps

dryabinin-actionengine:


ibgreen:
As discussed, try to make these changes in the common examples/webpack config .

Note that you can also add a separate rule for ts, rather than replacing modifying the existing rule.

ibgreen:
Also do not hesitate to add comments in the webpack config explaining why you are doing various changes.

belom88:
Only warnings but we need to fix them as well:
image

PR #1479, opened 2021-06-10
Fix for local webpack config

dryabinin-actionengine:
Checked the following examples by` yarn start-local`:

- Basis - OK
- Benchmark - ERROR
- 3d-tiles-with-cesium - ERROR Cesium is not defined (looks like an error not related to webpack config)
- 3d-tiles-with-three - OK
- gltf-with-deck.gl - OK
- gltf-with-raw-webgl - OK
- i3s-17-and-debug - OK
- 3d-tiles -OK
- GLTF -OK
- i3s -OK
- i3s-acrgis - OK
- pointcloud - OK
- textures - OK

### DIFF
--- a/examples/webpack.config.local.js
+++ b/examples/webpack.config.local.js
@@ -73,6 +73,7 @@ const LOCAL_DEVELOPMENT_CONFIG = {
   },
 
   resolve: {
+    extensions: ['.ts', '.tsx', '.js', '.json'],
     // mainFields: ['esnext', 'browser', 'module', 'main'],
     // Imports the library from its src directory in this repo
     alias: Object.assign({}, ALIASES)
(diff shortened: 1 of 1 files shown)


dryabinin-actionengine:
@ibgreen Benchmark example error:
image

examples/webpack.config.local.js
```diff
@@ -73,13 +73,23 @@ const LOCAL_DEVELOPMENT_CONFIG = {
...
     // mainFields: ['esnext', 'browser', 'module', 'main'],
     // Imports the library from its src directory in this repo
     alias: Object.assign({}, ALIASES)
   },
 
   module: {
     rules: [
+      {
+        // Transpile ES6 to ES5 with babel
+        test: /\.js|ts$/,
```
ibgreen, line 86: Would it be simpler to just add a new rule for `.ts`, rather than modify the rule for .js?

ibgreen (approved):

ibgreen:
Let's land this and deal with the benchmark break. You could also try to comment out the JSON benchmarks if only those are breaking.

PR #1485, opened 2021-06-11
Refactoring in webpack-config-local

dryabinin-actionengine:
Now benchmark app works in local mode


dryabinin-actionengine:
@ibgreen Please take a look at this. I think we should merge it if you don't mind.

ibgreen (approved):
Perfect

PR #1504, opened 2021-06-15
Restore functionality of passing options from gltf to draco loader

Avnerus:
Hi!
Seems that in the transition of the glTF loader to typescript, the ability to pass options from the glTF loader to the draco loader (such as `maxConcurrecny` or `draco` options) was lost. The reduced options object also evoked warnings about missing top-level options. This PR restores the lost functionality.

cc: @ibgreen


ibgreen (commented):
@Avnerus This fixes the options propagation but kills the types. 

I took an alternative approach in #1509 . Please test and rebase this PR if issues remain, if not close.

PR #1505, opened 2021-06-15
Remove segmentation data property from i3s example

dryabinin-actionengine:


ibgreen (approved):

examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
```diff
@@ -0,0 +1,117 @@
+import GL from '@luma.gl/constants';
```
belom88, line 1: This file should be removed

belom88 (changes requested):
Looks like wrong conflict resolving. `tile-lalyer.js` has to be deleted.

examples/experimental/i3s-17-and-debug/tile-layer/tile-layer.js
dryabinin-actionengine: Yes sure. Already removed.

PR #1517, opened 2021-06-17
Remove hardcoded dev worker config

dryabinin-actionengine:
Just remove hardcode of dev env. 
@ibgreen @Avnerus Maybe I missed some hidden issue related to my changes.
Do you know any?

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.2",
     "@babel/plugin-proposal-optional-chaining": "^7.14.2",
+    "@babel/plugin-transform-runtime": "^7.14.5",
     "@babel/register": "^7.13.16",
     "@luma.gl/core": "^8.3.0",
     "@luma.gl/debug": "^8.3.0",
--- a/scripts/webpack/worker.js
+++ b/scripts/webpack/worker.js
@@ -113,7 +113,7 @@ function addESNextSettings(config) {
 
 module.exports = (env = {}) => {
   let config = CONFIG;
-  env.dev = true;
+
   if (env.dev) {
     config.mode = 'development';
     config = addESNextSettings(config);
--- a/scripts/webpack/worker.js
+++ b/scripts/webpack/worker.js
@@ -14,6 +14,7 @@ const BABEL_CONFIG = {
     ['@babel/preset-env', {modules: false}]
   ],
   plugins: [
+    '@babel/plugin-transform-runtime',
     // webpack 4 cannot parse the most recent JS syntax
     '@babel/plugin-proposal-optional-chaining',
     '@babel/plugin-proposal-nullish-coalescing-operator',
--- a/yarn.lock
+++ b/yarn.lock
@@ -758,7 +758,7 @@
   dependencies:
     "@babel/helper-plugin-utils" "^7.14.5"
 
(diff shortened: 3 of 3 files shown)


dryabinin-actionengine:
@ibgreen Hmm I see failed pipeline [here](https://github.com/visgl/loaders.gl/pull/1517/checks?check_run_id=2846665527)

Avnerus:
Perhaps adding [this plugin](https://babeljs.io/docs/en/babel-plugin-transform-runtime) to the production build would help?

ibgreen:
That plugin definitely should be part of an es5 build, worth checking

dryabinin-actionengine:
Added `@babel/plugin-transform-runtime` plugin. Now pipeline passed successfully.

Avnerus:
Hi, this looks good.
But now I noticed that in Draco the `build-worker-dev` script is gone from `package.json` after the move to Typescript.  @ibgreen is there a reason for that?

ibgreen (approved):

ibgreen:
> I noticed that in Draco the build-worker-dev script is gone from package.json after the move to Typescript.

@Avnerus Can you confirm what is missing in https://github.com/visgl/loaders.gl/blob/master/modules/draco/package.json#L35-L37

Avnerus:
> > I noticed that in Draco the build-worker-dev script is gone from package.json after the move to Typescript.
> 
> @Avnerus Can you confirm what is missing in https://github.com/visgl/loaders.gl/blob/master/modules/draco/package.json#L35-L37

Hi, I was referring to [this line](https://github.com/visgl/loaders.gl/blob/c1f7d7bec911c704a190dbcf6b392a07a555771b/modules/draco/package.json#L37) that went missing.

ibgreen:
We are building minified workers with source maps, so there shouldn't be a need for a separate dev build.
- https://unpkg.com/@loaders.gl/i3s@3.0.0-alpha.21/dist/i3s-content-worker.js
- https://unpkg.com/@loaders.gl/i3s@3.0.0-alpha.21/dist/i3s-content-worker.js.map

Are the source maps not being loaded for you when breaking in the worker? I just tested and it seemed to work...

ibgreen:
I also noticed that we are loading worker via two different tags, that may be related to the open versioning bug...

Screen Shot 2021-06-18 at 5 48 49 PM

ibgreen:
@belom88 @dryabinin-actionengine It is also worrisome that we are loading the worker many times when we only create 3 workers in each queue.

Avnerus:
> We are building minified workers with source maps, so there shouldn't be a need for a separate dev build.
> 
>     * https://unpkg.com/@loaders.gl/i3s@3.0.0-alpha.21/dist/i3s-content-worker.js
> 
>     * https://unpkg.com/@loaders.gl/i3s@3.0.0-alpha.21/dist/i3s-content-worker.js.map
> 
> 
> Are the source maps not being loaded for you when breaking in the worker? I just tested and it seemed to work...

I have used the dev worker to make realtime changes to the code, but I support it should be enough in most cases to just disable the worker altogether when developing. For me it's OK to drop the dev workers.

ibgreen:
> I have used the dev worker to make realtime changes to the code

Understood. I feel that the build times for the loaders repo have gotten quite out of hand, so I prefer to reduce the number of build artifacts unless there are strong reasons for keeping them. Let's drop it for now, we can always add it back later.

PR #1545, opened 2021-06-23
Convert i3s-tile-converter to typescript

dryabinin-actionengine:
It is first small step to convert i3s-tile-converter to typescript.
Next steps:
1. Remove d.ts file and move all types to i3s-converter.ts file.
2. Enable validation for i3s-converter.ts in ts config.


ibgreen (approved):
This looks great, thanks!

PR #1568, opened 2021-06-26
Remove tileset options that override fetch options

Avnerus:
Hi,
These options were causing unwanted behavior because the default `headers` was `{}`, thus overriding `fetch.headers` regardless of what is sent by the user.
I think it should be OK to just remove these options and require everything to be set at `{fetch: }`?
I kept `token` because I saw it was getting added to `queryParams`, probably i3s related.


ibgreen (approved):
Yes we should standardize on `loadOptions`. @belom88 @dryabinin-actionengine please land if OK.

belom88:
This is exactly what I've done in https://github.com/visgl/loaders.gl/pull/1550

PR #1572, opened 2021-06-28
I3S-App - Fix picking functionality

dryabinin-actionengine:
1. Fix for picking functionality
2. Enable 1.7 tileset by default

### DIFF
--- a/modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
+++ b/modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
@@ -67,6 +67,7 @@ export async function parseI3STileContent(
 }
 
 /* eslint-disable max-statements */
+// eslint-disable-next-line complexity
 async function parseI3SNodeGeometry(arrayBuffer: ArrayBuffer, tile: Tile = {}) {
   if (!tile.content) {
     return tile;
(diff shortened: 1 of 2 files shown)


modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
```diff
@@ -108,9 +109,19 @@ async function parseI3SNodeGeometry(arrayBuffer: ArrayBuffer, tile: Tile = {}) {
 
     for (const key in decompressedGeometry.loaderData.attributes) {
       const dracoAttribute = decompressedGeometry.loaderData.attributes[key];
-      if (dracoAttribute.name === 'POSITION') {
-        attributes.position.metadata = dracoAttribute.metadata;
-        break;
+
+      switch (dracoAttribute.name) {
+        case 'POSITION': {
```
belom88, line 114: You don't need `{` in case statement.

modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
```diff
@@ -108,9 +109,19 @@ async function parseI3SNodeGeometry(arrayBuffer: ArrayBuffer, tile: Tile = {}) {
...
+          attributes.position.metadata = dracoAttribute.metadata;
+          break;
+        }
+        case 'feature-index': {
+          attributes.id.metadata = dracoAttribute.metadata;
+          break;
+        }
+        default: {
+          break;
```
belom88, line 123: `break` in `default` doesn't have sense because `default` is last option.

belom88 (approved):
Ok with minor comments

modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
dryabinin-actionengine: EsLint ask for for something in default case.

modules/i3s/src/lib/parsers/parse-i3s-tile-content.ts
dryabinin-actionengine: Removed

PR #1577, opened 2021-06-28
Fix ArrowJS with large datasets Observable notebook link to point to …

RandomFractals:
…the original

the fork listed in your docs is broken and contributes nothing to the original notebook I created.


ibgreen (approved):
Thanks!

PR #1593, opened 2021-07-01
feat(core): filesystem improvements

ibgreen:
@kylebarron includes small steps toward range support.


ibgreen:
This code is already merged.

PR #1599, opened 2021-07-02
Fix draco parser

dryabinin-actionengine:
This changes fix the following items:
1. GLTF examples on the website
2. I3S-Converter Geometry


ibgreen (approved):
Nice work!

PR #1619, opened 2021-07-06
Refactor 3DTiles converter to typescript

dryabinin-actionengine:


ibgreen (approved):
This looks good.

As next step, I would create separate type files for 
- `modules/3d-tiles/.../3d-tiles-schema.ts`
- `modules/i3s/.../i3s-schema.ts`
- `modules/tiles/.../tiles-schema.ts`

The idea being that `3d-tiles` and `i3s` normalize their output into the `tiles` schema.

PR #1650, opened 2021-07-11
examples: i3s move to website folder

belom88:
- removed `examples/website/i3s`
- moved `examples/experimental/i3s-17-and-debug` -> `examples/website/i3s`
- updated `website/gatsby-config.js`


ibgreen (approved):
This is not just a move, there are a number of changed files, seems to be some cleanup too?

belom88:
> This is not just a move, there are a number of changed files, seems to be some cleanup too?

Not really. There was "i3s" folder already and there were files with same names

PR #1670, opened 2021-07-15
Bump math.gl to 3.5.1

felixpalmer:


dryabinin-actionengine (approved):

Pessimistress:
This PR is breaking downstream modules. All math.gl dependencies should be specified with `^` instead of locked to a specific version. The current production release blocks deck.gl and luma.gl from pruning their math dependencies.

dryabinin-actionengine:
> This PR is breaking downstream modules. All math.gl dependencies should be specified with `^` instead of locked to a specific version. The current production release blocks deck.gl and luma.gl from pruning their math dependencies.

Ok. I crated a [PR](https://github.com/visgl/loaders.gl/pull/1689) for that.

PR #1686, opened 2021-07-19
[parquet] fix internal dependency versions

Pessimistress:
Internal dependency (@loaders/*) versions must match the current package version, otherwise lerna does not bump them during publish.

### DIFF
--- a/modules/parquet/package.json
+++ b/modules/parquet/package.json
@@ -39,9 +39,9 @@
     "lzo": false
   },
   "dependencies": {
-    "@loaders.gl/compression": "3.0.0-beta.8",
-    "@loaders.gl/loader-utils": "3.0.0-beta.8",
-    "@loaders.gl/schema": "3.0.0-beta.8",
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
We are on 3.0.4

dryabinin-actionengine:
@Pessimistress @ibgreen  I guess it can be closed because of this [PR](https://github.com/visgl/loaders.gl/pull/1687).
More than that we are on `3.0.5` already.

Pessimistress:
Why is this closed? The master is on 3.0.1. If you don't fix the versions on master then when you publish 3.1.0-alpha the dependencies will still not be synced.

dryabinin-actionengine:
> Why is this closed? The master is on 3.0.1. If you don't fix the versions on master then when you publish 3.1.0-alpha the dependencies will still not be synced.

Yes, I see. I missed target branch with 3.0-release. Reopened

PR #1687, opened 2021-07-19
[parquet] fix internal dependency versions

Pessimistress:
Internal dependency (@loaders/*) versions must match the current package version, otherwise lerna does not bump them during publish.

### DIFF
--- a/modules/parquet/package.json
+++ b/modules/parquet/package.json
@@ -39,8 +39,8 @@
     "lzo": false
   },
   "dependencies": {
-    "@loaders.gl/loader-utils": "3.0.0-beta.1",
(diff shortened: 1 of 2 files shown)


ibgreen (approved):

dryabinin-actionengine:
@ibgreen @Pessimistress Looks like we should change dependency version to `3.0.5`

Pessimistress:
@dryabinin-actionengine You need to rebase 3.0-release branch and run yarn again. The lock file is still out of sync.

belom88:
This branch was rebased and merged into 3.0-release by @dryabinin-actionengine in https://github.com/visgl/loaders.gl/pull/1690

PR #1703, opened 2021-07-26
fix(3d-tiles): cesium rtc parse

belom88:


ibgreen (approved):
Interesting... do we know why this broke?

Is this fixing a regression due to some change in parsers?

has this never worked before? That would be surprising...

belom88:
> Interesting... do we know why this broke?
> 
> Is this fixing a regression due to some change in parsers?
> 
> has this never worked before? That would be surprising...

Screenshot from 2021-07-26 17-48-10

belom88:
I guess we need this in the next patch...

PR #1711, opened 2021-07-27
fix: wrongly defined variable type

zuiidea:
cannot assign to "classIds" because it is a constant


belom88 (approved):
Interestingly, why ts doesn't fail `=` operators of `const`. It should check this.

belom88:
I see. `/* eslint-disable */` in the top of the file.

belom88:
This fix is delivered in https://github.com/visgl/loaders.gl/pull/1714

PR #1712, opened 2021-07-27
Bump deck.gl to 8.5.0 and luma.gl to 8.5.4

dryabinin-actionengine:
We should bump luma.gl version as well to avoid version conflicts between loaders.gl and deck.gl.


dryabinin-actionengine:
@ibgreen @belom88 Looks like I3S examples breaks after these changes.

ibgreen (approved):
The breakage is independent of this PR.

PR #1722, opened 2021-07-29
Add Parquet worker

dryabinin-actionengine:


ibgreen (approved):
Thanks.

For this worker to be really efficient, we need to send back data in binary (columnar) form, not rows.

That is why it would be good to make the loader return e.g. a real arrow table.

PR #1724, opened 2021-07-29
MVTLoader binary mode supports feature.id

Pessimistress:
For https://github.com/visgl/deck.gl/issues/6006

- Add optional `fields` to the binary format
- MVTLoader saves top-level `id` field if present
- `binaryToGeojson` restores extra fields to features
- Test


modules/gis/docs/api-reference/geojson-to-binary.md
```diff
@@ -60,7 +60,9 @@ corresponds to 3D coordinates, where each vertex is defined by three numbers.
         numericProperty1: {value: Float32Array, size: 1}
     }
     // Array of objects with non-numeric properties from Point geometries
-    properties: [{PointFeatureProperties}]
+    properties: [{PointFeatureProperties}],
+    // Non-standard top-level fields
+    fields: [{PointFeatureExtraFields}]
```
kylebarron, line 65: It might be good to mention that for now this just includes `id`?

kylebarron (approved):

PR #1728, opened 2021-07-30
Parquet - Snappy compression fix

dryabinin-actionengine:


ibgreen (approved):
Makes sense - I noticed that we were sending in empty strings to snappy decompressor.

PR #1730, opened 2021-08-02
chore(wkt): improved wkt typings

Evin-K:
added more accurate types


belom88 (commented):
Adhere to https://www.conventionalcommits.org/en/v1.0.0/ for commit naming

modules/wkt/src/lib/encode-wkt.ts
```diff
@@ -1,23 +1,25 @@
...
  */
 
 export default function encodeWKT(gj: {
   type: string;
   properties?: {};
-  geometry?: any;
+  geometry?: {
+    type: string;
+    coordinates: any[];
+  };
```
kylebarron, line 15: Should we use `@types/geojson` typing for `geometry` and `geometries`?

modules/wkt/src/lib/encode-wkt.ts
```diff
@@ -38,24 +40,24 @@ export default function encodeWKT(gj: {
...
+      return `MULTIPOINT (${ringWKT(gj.coordinates as number[][])})`;
     case 'MultiPolygon':
-      return 'MULTIPOLYGON (' + multiRingsWKT(gj.coordinates as any[]) + ')';
+      return `MULTIPOLYGON (${multiRingsWKT(gj.coordinates as number[][][][])})`;
     case 'MultiLineString':
-      return 'MULTILINESTRING (' + ringsWKT(gj.coordinates as any[]) + ')';
+      return `MULTILINESTRING (${ringsWKT(gj.coordinates as number[][][])})`;
     case 'GeometryCollection':
-      return 'GEOMETRYCOLLECTION (' + gj.geometries.map(encodeWKT).join(', ') + ')';
+      return `GEOMETRYCOLLECTION (${gj.geometries!.map(encodeWKT).join(', ')})`;
```
kylebarron, line 60: If you use the `geometry` type from `@types/geojson`, I don't think you'd need any of these casts.

kylebarron: Another benefit to using the `Geometry` type is that you can assert `: never` after this last case to make it clear that we're doing all exhaustive checks for a `Geometry` object

modules/wkt/src/lib/encode-wkt.ts
```diff
@@ -38,24 +40,24 @@ export default function encodeWKT(gj: {
   }
 
   function wrapParens(s: string) {
-    return '(' + s + ')';
+    return `(${s})`;
```
kylebarron, line 43: Not specifically related to this PR but is there a performance hit to defining these helper functions within this function instead of in higher scope? If you had 100,000 geometries that you wanted to encode to WKT, and did `geometries.map(encodeWKT)`, then each of these helper functions would be defined 100,000 times, instead of once? 

It looks like maybe this isn't an issue for modern browsers? https://stackoverflow.com/a/10204449 cc @ibgreen ?

modules/wkt/src/lib/encode-wkt.ts
```diff
@@ -38,24 +40,24 @@ export default function encodeWKT(gj: {
...
   }
 
   switch (gj.type) {
     case 'Point':
-      return 'POINT (' + pairWKT(gj.coordinates as any[]) + ')';
+      return `POINT (${pairWKT(gj.coordinates as number[])})`;
```
kylebarron, line 48: Nit: Since we've already defined a `wrapParens` helper, it might be clearer to use that? Right now since the expression inside the `${}` is a function, it's a little hard to tell that the `)` right before the closing `` ` `` is part of the string, and not the expression. Might be clearest IMO to have
```js
return 'POINT ' + wrapParens(pairWKT(gj.coordinates));
```

modules/wkt/src/lib/encode-wkt.ts
kylebarron: So the input type can be `gj: Feature | Geometry` where both are imported from `@types/geojson`

kylebarron (approved):

dryabinin-actionengine (approved):

PR #1738, opened 2021-08-04
chore(tsc): add noUnusedLocals rule to tsconfig

igorDykhta:
- add noUnusedLocals rule to tsconfig
- // @ts-ignore in all places to pass noUnusedLocals rule


ibgreen (approved):
Are the some cases where we could comment out / remove instead of adding ts-ignore?

PR #1748, opened 2021-08-09
bug(tiles): pass user-defined options

belom88:


belom88:
Another workaround was found

PR #1754, opened 2021-08-10
Add part of dummy header to LZ4 parquet files

dryabinin-actionengine:
Parquet files with LZ4 compression doesn't have "header" with magic number + frame descriptor part.
The idea is to add dummy header to file buffer.
[Spec](https://android.googlesource.com/platform/external/lz4/+/HEAD/doc/lz4_Frame_format.md)
Here an [example](https://github.com/Benzinga/lz4js/blob/master/lz4.js#L474) of encoding lz4 file from package which we use to decompress such files.

### DIFF
--- a/modules/compression/src/lib/lz4-compression.ts
+++ b/modules/compression/src/lib/lz4-compression.ts
@@ -1,9 +1,16 @@
 // LZ4
+import {concatenateArrayBuffers} from '@loaders.gl/loader-utils/';
 import type {CompressionOptions} from './compression';
 import {Compression} from './compression';
 // import lz4js from 'lz4js'; // https://bundlephobia.com/package/lz4
 
 let lz4js;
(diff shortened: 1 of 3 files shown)


ibgreen (approved):
OK. this makes sense. Of course, without a passing test case this doesn't give us much confidence.

Also, could we package the header generation inside a new `LZ4RawCompression` class, instead of adding the code to `compression.ts`?

dryabinin-actionengine:
> OK. this makes sense. Of course, without a passing test case this doesn't give us much confidence.
> 
> Also, could we package the header generation inside a new `LZ4RawCompression` class, instead of adding the code to `compression.ts`?

Yes sure, but first of all we need fully implement LZ4 heading. For now it is just one more step forward. It doesn't work now.

PR #1766, opened 2021-08-18
DO NOT MERGE: Add LZ4 compression support for Parquet

dryabinin-actionengine:
Looks like current `lz4js` library can't handle lz4 compressed blocks without lz4 header in parquet file properly.
I just tried to use lz4 module which is suggested in [Apache Parquet repo.](https://github.com/apache/parquet-format/blob/master/Compression.md)
The weakness of LZ4 module that it consumes only Node.js `Buffer` which is not really good if we want to support Parquet loader in browser.
Next step is to compare [lz4js](https://github.com/Benzinga/lz4js) and [lz4](https://github.com/pierrec/node-lz4) realisations and try to make `lz4js` work like `lz4` does.

### DIFF
--- a/modules/compression/package.json
+++ b/modules/compression/package.json
@@ -28,6 +28,7 @@
   ],
   "browser": {
     "lzo": false,
+    "lz4": false,
     "zlib": false,
     "util": false
   },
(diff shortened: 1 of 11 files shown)


dryabinin-actionengine:
Closed because of https://github.com/visgl/loaders.gl/pull/1769

PR #1775, opened 2021-08-20
chore(I3S-Debug): add tooltips

morphoix:


examples/website/i3s/components/tool-bar.js
```diff
@@ -7,19 +7,18 @@ import { faMap, faBug, faSdCard, faExclamationCircle, faInfo, faGlobe } from '@f
...
+  bottom: 0;
```
dryabinin-actionengine, line 13: Remove empty spaces.

examples/website/i3s/components/tool-bar.js
```diff
@@ -29,21 +28,22 @@ const Container = styled.div`
...
-  min-width: 40px;
-  flex: 1 0 1px;
-  padding: 0;
+  flex: 1 1 0px;
+  margin: 2px 0 2px 1px;
+  height: 56px;
+  width: 16.5vw;
   &:hover {
     transition: all 1s;
     color: white;
```
dryabinin-actionengine, line 40: To avoid hover effect on mobile while click on button.
Also animation is removed.
```suggestion
  @media (hover: hover) and (pointer: fine) {
     &:hover { color: white; }
   }
```

dryabinin-actionengine (changes requested):
Also I found that in landscape mode on the Iphone x for example we have frozen hover effect if we click on tool bar button. To prevent such behaviour I suggest to use special @media rule.

examples/website/i3s/components/tool-bar.js
```diff
@@ -66,6 +66,63 @@ const Title = styled.h3`
...
+  font-weight: 500;
+  font-size: 16px;
+  visibility: hidden;
+  color: transparent;
+  background-color: transparent;
+  text-align: center;
+  width: 110px;
+  line-height: 22px;
+  border-radius: 4px;
+  transition: visibility 0.5s, color 0.5s, background-color 0.5s, width 0.5s ease-in-out;
```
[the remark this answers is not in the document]
morphoix: yes, my Lord

PR #1787, opened 2021-08-24
chore(textures): remove unnecessary dependencies

### LINKED ISSUE #1785: textures module requires loaders.gl/core alpha version
Reported by @trxcllnt 

It is just a devDependency - can we remove it?

loader modules should not depend on core, but I seem to remember that there was a complication here..

morphoix:
Fixes #1785

### DIFF
--- a/modules/geotiff/package.json
+++ b/modules/geotiff/package.json
@@ -39,8 +39,5 @@
     "@loaders.gl/worker-utils": "3.0.0-alpha.18",
     "fast-xml-parser": "^3.16.0",
     "geotiff": "^1.0.4"
-  },
-  "devDependencies": {
-    "@loaders.gl/core": "3.0.0-alpha.18"
   }
 }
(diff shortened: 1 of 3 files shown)


ibgreen (approved):
Thanks for the quick PR. Note that per comments in the issue #1785  there are at least one more place where an alpha release is referenced.

morphoix:
> Thanks for the quick PR. Note that per comments in the issue #1785 there are at least one more place where an alpha release is referenced.

ok, fixed

dryabinin-actionengine (approved):

modules/geotiff/package.json
```diff
@@ -39,8 +39,5 @@
     "@loaders.gl/worker-utils": "3.0.0-alpha.18",
```
Pessimistress, line 39: Still not matching package version

modules/zarr/package.json
```diff
@@ -32,8 +32,5 @@
   "dependencies": {
     "@loaders.gl/worker-utils": "3.0.0-alpha.18",
```
Pessimistress, line 33: Ditto

PR #1795, opened 2021-08-24
bug(schema): fixed Can't reexport the named export

belom88:


ibgreen (commented):
I am worried about this fix. I think this is importing from different dists. Mixing dists is a recipe for pain. 

@trxcllnt

belom88:
`yarn start-local` doesn't work with default dist

trxcllnt:
`Arrow.node.ts` and `Arrow.dom.ts` [re-export from `Arrow.ts`](https://github.com/apache/arrow/blob/04575bc8e90216d8e685d1b857f487708426001c/js/src/Arrow.dom.ts#L113), so this should be fine.

ibgreen (approved):
> Arrow.node.ts and Arrow.dom.ts re-export from Arrow.ts, so this should be fine.

OK. I admit that I still quietly wonder why changing this fixes anything?

And it looks odd to use two different ways of importing next to each other.

I expect that applications will still import for `apache-arrow`.

trxcllnt:
The only thing that comes to mind is there may be a bug with the bundler used here. There should be no functional difference between importing from `apache-arrow` and `apache-arrow/Arrow` (aside from the additional DOM or node-specific exports).

PR #1809, opened 2021-08-25
chore(parquet): refactor parse function

belom88:
I found out that converting to Blob is not optimal choice because `parquestjs` converts it back to `ArrayBuffer` internally.

I made a test for `parseFileInBatches`. @ibgreen, did you suppose to  create some global function `parseFileInBatches` in `core` module?


belom88:
Well, `parseFileInBatches` is calling `openBlob` in this PR.
The change with `openArrayBuffer` is only for `parse` function.
If it is not good for upcoming design changes then I can change `parse` back on `openBlob`. But this PR also changes place of `yield` call. It is the better place in my opinion.

PR #1841, opened 2021-09-15
fix(i3s): change logic of tileset urls parsing for i3s apps

dryabinin-actionengine:
Now I**3S App** and **I3S-Debug App** are able to handle the following tileset urls:

1. /?url=https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer
2.  /?url=https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer/
3.  /?url=https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer/layers/0
3.  /?url=https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer/layers/0/

image

### DIFF
--- a/examples/website/i3s/url-utils.js
+++ b/examples/website/i3s/url-utils.js
@@ -20,7 +20,8 @@ export function parseTilesetUrlParams(url, options) {
 
 function prepareTilesetUrl(parsedUrl) {
   if (parsedUrl.pathname.includes('layers/0')) {
-    return parsedUrl.href;
+    // Replace useless last "/" character if it presents.
+    return parsedUrl.href.replace(/\/+$/, '');
   }
   // Add '/' to url if needed + layers/0 if not exists in url.
   const replacedPathName = parsedUrl.pathname.replace(/\/?$/, '/').concat('layers/0');
--- a/examples/website/i3s/url-utils.js
+++ b/examples/website/i3s/url-utils.js
@@ -19,7 +19,10 @@ export function parseTilesetUrlParams(url, options) {
 }
 
 function prepareTilesetUrl(parsedUrl) {
-  if (parsedUrl.pathname.includes('layers/0')) {
+  // Try to find particular layer in url.
+  const layer = parsedUrl.pathname.match(/layers\/\d/);
+
+  if (layer) {
     // Replace useless last "/" character if it presents.
     return parsedUrl.href.replace(/\/+$/, '');
   }


ibgreen (approved):

examples/website/i3s/url-utils.js
```diff
@@ -20,7 +20,8 @@ export function parseTilesetUrlParams(url, options) {
 
 function prepareTilesetUrl(parsedUrl) {
   if (parsedUrl.pathname.includes('layers/0')) {
```
Tamrat-B, line 22: why hardcode to 'layers/0' ? If I have a service like this '.. rest/services/SomeService/SceneServer/layers/23' (which is possible) this wouldn't' work. The layer id is derived from SceneServer.layers[0].id; but while parsing we don't need to know what the value is, we should just be literal and use it.

Tamrat-B (changes requested):

examples/website/i3s/url-utils.js
dryabinin-actionengine: Thank you for the comment! I fixed regular expression to handle any digit value after `layers/`. 
What logic should be implemented if user just types `https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer` without particular layer? Should we use `layers/0` by default?

Tamrat-B: when the user specifies just only `/SceneServer/`, get the json resource and grep for the `layers` array. Loop thru all members of the `layers` array and add them to the scene, by appending the `href` value from `layers[`layerID`].href` resource. The `href` property is normative so it should be present. This `href `is composed of the named rest resource `layers/` + `layers[count-1].id` resulting in something like this: "./layers/0" You can see in this example as well: https://tiles.arcgis.com/tiles/u0sSNqDXr7puKJrF/arcgis/rest/services/Frankfurt2017_v17/SceneServer.

Tamrat-B (approved):

examples/website/i3s/url-utils.js
dryabinin-actionengine: Cool. Thank you!
We should implement such behaviour. I'v created an [issue](https://github.com/visgl/loaders.gl/issues/1846) for that.
Now we don't have such logic in loaders.gl. We just use` layers/0` if nothing is provided after `SceneServer/`

PR #1853, opened 2021-09-24
Consistently use core as peer dependency

Pessimistress:
Having multiple copies of core is bad.


ibgreen (approved):
Thanks. The loader modules should not depend on core at all, but until that is addressed this makes sense.

PR #1854, opened 2021-09-26
feat(examples): i3s - building explorer ui

belom88:
image

### DIFF
--- a/examples/website/i3s/app.js
+++ b/examples/website/i3s/app.js
@@ -232,11 +232,11 @@ export default class App extends PureComponent {
     const style = {
       display: 'flex',
       top: '125px'
-    }
+    };
     if (!showMemory) {
       style.display = 'none';
     }
-    if (sublayers?.length) {
+    if (sublayers.length) {
       style.top = '159px';
     }
     return <StatsWidgetContainer style={style} ref={(_) => (this._statsWidgetContainer = _)} />;
--- a/examples/website/i3s/app.js
+++ b/examples/website/i3s/app.js
@@ -17,8 +17,9 @@ import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import {faSpinner} from '@fortawesome/free-solid-svg-icons';
 import {INITIAL_EXAMPLE_NAME, EXAMPLES} from './examples';
 import {INITIAL_MAP_STYLE} from './constants';
-import { Color, Flex, Font } from './components/styles';
+import {Color, Flex, Font} from './components/styles';
 import {load} from '@loaders.gl/core';
+import {buildSublayersTree} from './helpers/sublayers';
 
 const TRANSITION_DURAITON = 4000;
 const BUILDING_SCENE_LAYER_TYPE = 'Building';
@@ -54,7 +55,7 @@ const StatsWidgetContainer = styled.div`
   padding: 24px;
   border-radius: 8px;
   line-height: 135%;
-  top: 125px;
+  top: 159px;
   bottom: auto;
   width: 277px;
   left: 10px;
@@ -77,10 +78,12 @@ export default class App extends PureComponent {
       selectedFeatureIndex: -1,
       isAttributesLoading: false,
       showMemory: true,
-      layerUrls: []
+      layerUrls: [],
+      sublayers: []
     };
     this._onSelectTileset = this._onSelectTileset.bind(this);
     this.handleClosePanel = this.handleClosePanel.bind(this);
+    this._setMemoryVisibility = this._setMemoryVisibility.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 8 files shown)


examples/website/i3s/app.js
```diff
@@ -214,21 +227,33 @@ export default class App extends PureComponent {
...
     // TODO - too verbose, get more default styling from stats widget?
-    return <StatsWidgetContainer ref={(_) => (this._statsWidgetContainer = _)} />;
+    const style = {
+      display: 'flex',
+      top: '125px'
+    }
+    if (!showMemory) {
+      style.display = 'none';
+    }
+    if (sublayers?.length) {
```
ibgreen, line 239: initialize sublayers with empty array so we don't need the `?.` ?

examples/website/i3s/helpers/sublayers.js
```diff
@@ -0,0 +1,36 @@
+export function buildSublayersTree(sublayers) {
+  const fullModel = sublayers.find((subtree) => subtree.modelName === 'FullModel');
+  const result = postprocessNode(fullModel);
+  return result;
+}
+
+function postprocessNode(sublayer) {
```
ibgreen, line 7: Don't you feel like this type of functionality should be provided by the `@loaders.gl/i3s` module? A good "litmus test" is to if anyone using the building layers would need to write this code...

ibgreen (approved):

examples/website/i3s/app.js
```diff
@@ -178,20 +190,21 @@ export default class App extends PureComponent {
...
-        onTilesetLoad: this._onTilesetLoad.bind(this),
-        onTileLoad: () => this._updateStatWidgets(),
-        onTileUnload: () => this._updateStatWidgets(),
-        // TODO enable it when Building Scene Layer picking will be implemented.
-        pickable: layerType !== BUILDING_SCENE_LAYER_TYPE,
-        loadOptions,
-        highlightedObjectIndex: selectedFeatureIndex
-      })
-    ));
+    return layerUrls.map(
```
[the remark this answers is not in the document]
belom88: Each sublayer is an independent layer with type "3DObjects".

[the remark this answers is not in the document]
ibgreen: Yes I understand. At least one concerns is that they all have their own request schedulers. The job of the request scheduler is partially to make sure we don't issue more requests than the browser can process in parallel. This is defeated when we don't share a scheduler between all these tilesets. It may impact performance when panning and zooming as a lot more requests that are no longer needed will be queued up.

examples/website/i3s/helpers/sublayers.js
belom88: I mostly created this code to 
1. deep copy the tree for React
2. filter "preview" layer
3. calculate the number of child nodes for each node (just in case)
It is not necessary for general users of I3SBuildingSceneLayerLoader. For somebody who wants to show "preview" layer it even would be breaking thing

ibgreen: Understood. Some JSDoc headers with short explanations would probably be helpful. After all this is an example that should be easy for users to understand.

examples/website/i3s/components/building-explorer.js
```diff
@@ -0,0 +1,161 @@
...
+        </CheckboxContainer>
+      );
+    });
+  }
+
+  render() {
+    const {sublayers, setMemoryVisibility} = this.props;
+    const {showExplorerTree} = this.state;
+    return (
+      <BuildingExplorerContainer style={{height: showExplorerTree ? undefined : '36px'}}>
```
dryabinin-actionengine, line 139: You can use props in styled components just pass `showExplorerTree` as attribute here and get it from styled component block as props => props.showExplorerTree ? 36px : auto. Take a look how it is done in other places.

examples/website/i3s/app.js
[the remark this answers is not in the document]
belom88: I see. Now data flows though deck.gl. The tileset object (with scheduler) is created in deck.gl. We need to think about all this interaction between loaders.gl and deck.gl

examples/website/i3s/app.js
```diff
@@ -109,16 +112,25 @@ export default class App extends PureComponent {
...
-   * @param {string} tilesetUrl 
+   * @param {string} tilesetUrl
    * @returns {string[]} Sublayer urls or tileset url.
    * TODO Add filtration mode for sublayers which were selected by user.
    */
   async getLayerUrls(tilesetUrl) {
     try {
       const tileset = await load(tilesetUrl, I3SBuildingSceneLayerLoader);
-      return tileset?.sublayers.map(sublayer => sublayer.url);
+      const sublayersTree = buildSublayersTree(tileset.header.sublayers);
```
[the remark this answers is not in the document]
belom88: See https://github.com/visgl/loaders.gl/pull/1854#discussion_r716399511

examples/website/i3s/app.js
belom88: Done

examples/website/i3s/helpers/sublayers.js
belom88: Done

examples/website/i3s/components/building-explorer.js
belom88: Done

morphoix (approved):

dryabinin-actionengine:
Good, One minor thing I would remove this list dots from subcategories. 
image

examples/website/i3s/app.js
```diff
@@ -109,16 +112,25 @@ export default class App extends PureComponent {
...
+      this.setState({sublayers: sublayersTree.sublayers});
+      const urls = tileset?.sublayers
+        .map((sublayer) => sublayer.url)
+        .filter(
+          (url) =>
+            url !==
+            'https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/Admin_Building_v17/SceneServer/layers/0/sublayers/0'
```
dryabinin-actionengine, line 131: I am not sure that we should save this line.

belom88: We should. It is "Preview" model. We shouldn't show "Preview" and "Full Model" at the same time

belom88:
> Good, One minor thing I would remove this list dots from subcategories.
> image

Dots are placeholders. It was the easiest way to make placeholder. We can use any icon from FontAwesome though

PR #1902, opened 2021-10-28
fix(tile-converter): i3s->3dtiles - model matrix transformation

belom88:


ibgreen (approved):
Since we can specify a modelMatrix to deck.gl (which then gets applied on the GPU), it is not fully clear to me why we need to deal with it on the CPU, but apart from a performance hit, I don't have any concerns.

belom88:
I3S stores coordinates in lonlat offsets when 3DTiles - in meter offsets. Transformation matrix is not sufficient for this conversion.
We convert coordinates via pipeline:
(1) lonlat offsets -> (2)lonlat decimal degrees -(cartographicToCartesian)-> (3) cartesian -> (4) meter offsets

PR #1919, opened 2021-11-06
chrore(test-apps): remove husky

belom88:
husky in "typescript-test" application sets commit hooks that is not good for test application.


ibgreen (approved):
Feel free to simplify the test application - as long as it generates the TS errors.

PR #1974, opened 2021-11-30
fix(tile-converter): add build script for docker image of tile-converter

dryabinin-actionengine:
Docker doesn't now about **time**  command inside `time npm run tsbuild ` script.


ibgreen (approved):
Let's can just remove the `time` statements and keep a single target. I have debugged enough for now :)

PR #1975, opened 2021-12-01
fix(tile-converter): change converter options

dryabinin-actionengine:
1. Rename `--generate-texture` -> `--generate-textures`
2. Fix positions transformation.


belom88 (approved):

belom88 (changes requested):
Tests

ibgreen (approved):

PR #1981, opened 2021-12-02
chore(tile converter): refactoring

belom88:
- use @math.gl/geoid;
- use scratch vector during transform positions;
- fix test


ibgreen (approved):
Very nice!

PR #1982, opened 2021-12-02
test: switch converter tests on

belom88:
I picked a number of commits from `master` branch and found out that the converter tests are switched off on the `3.1-release` branch


ibgreen (approved):
Lots of yarn.lock changes where they intentional?

belom88:
> Lots of yarn.lock changes where they intentional?

It is because of lots of new commits with new dependencies

PR #1991, opened 2021-12-05
fix(tiles): culling relative to viewport center

belom88:


Pessimistress (approved):

ibgreen (approved):
This is for elevated viewpoints?

belom88:
Yes. Culling works wrong if viewport has elevation

PR #2039, opened 2021-12-22
[fix] selecLoader should consider baseUri of the data source

Pessimistress:


ibgreen (approved):

Pessimistress:
Seems like the change has some unintended impact on the GLTF loader. For patch purpose maybe a more localized fix is the way to go? #2040

PR #2046, opened 2021-12-30
fix(tiles): init zoom calculation

belom88:
Projection meaning:
image
`projection = zmin * tan(pitch)`

Calculate longitude offset and latitude offset having distance and bearing angle:
image
`[longitude offset] = projection * sin(bearing)`
`[latitude offset] = projection * cos(bearing)`

### DIFF
--- a/examples/website/i3s/app-debug.js
+++ b/examples/website/i3s/app-debug.js
@@ -60,7 +60,7 @@ import {
 import {Color, Flex, Font} from './components/styles';
 import {buildSublayersTree} from './helpers/sublayers';
 import {initStats, sumTilesetsStats} from './helpers/stats';
-import { getElevationByCentralTile } from './helpers/terrain-elevation';
+import {getElevationByCentralTile} from './helpers/terrain-elevation';
 
 const TRANSITION_DURAITON = 4000;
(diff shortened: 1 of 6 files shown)


ibgreen (commented):
A bit hard to review, no description, no comments stating what this is trying to accomplish.

Pessimistress:
This is not the right direction of solve the camera problem.

If the view state system does not support looking at an above-ground target, then we need to fix that system. Calculating a hypothetical lookAt point on the ground will not give you the right user interaction.

belom88:
It is just example apps. We can use some elevation prop in the viewState as soon as it is implemented in @deck.gl

Pessimistress:
Example apps are not for debugging. Users rely on them to understand how this library is supposed to be used. If it requires this much code to get it work right then there is something very wrong with what we offer.

belom88:
Am I correct that we need to add  some `elevation` prop to `Viewport` class to calculate `view` matrix properly for this case?

ibgreen (approved):
Good discussion.
- We definitely need a proper fix in deck.gl, and this fix should not go into any of our clean examples.
- However, this application is not a demo. It is really a showcase. The code is already way too complex and the purpose is to provide a tileset debug app for users not show how to use the libraries.

Issue #2047, opened 2022-01-06
tile-converter is storing duplicated textures when converting 3dtiles to i3s slpk bloating slpk size


Tamrat-B:
Texture resources are being duplicated when converting 3dtiles to i3s slpk. 

In addition to the correct texture output in the slpk, there is a duplicated texture (0\index.jpg and/or 1\index.ktx2) that should only be output when creating content for serving i3s (not in slpk).

belom88:
Converted slpk has file structure:
![image](https://user-images.githubusercontent.com/18549629/149013716-a1f80213-438e-4d67-acf1-18d23d9aa24a.png)
Expected result:
![image](https://user-images.githubusercontent.com/18549629/149013799-0593723a-4cca-441f-8ed2-38ca6c0e5993.png)

belom88:
spec: https://github.com/Esri/i3s-spec/blob/master/format/Indexed%203d%20Scene%20Layer%20Format%20Specification.md#example-18-slpk-structure-summary-for-3d-objects-

belom88:
3.1.5 has been published

PR #2048, opened 2022-01-06
fixed duplicated texture output

Tamrat-B:
@belom88 @dryabinin-actionengine Fix for #2047

### DIFF
--- a/modules/tile-converter/src/i3s-converter/i3s-converter.ts
+++ b/modules/tile-converter/src/i3s-converter/i3s-converter.ts
@@ -942,9 +942,6 @@ export default class I3SConverter {
     childPath: string,
     slpkChildPath: string
   ): Promise<void> {
-    const texturePath = join(childPath, `textures/${name}/`);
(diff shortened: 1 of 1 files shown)


belom88 (changes requested):
There should be logic of changing `textureSetDefinitions`

belom88:
@Tamrat-B , we will help you to land this PR

ibgreen (approved):

Tamrat-B:
> There should be logic of changing `textureSetDefinitions`

I think this particular issue was just a bug - there shouldn't be duplicate textures in the slpk. We should however, have an option with the converter to output only ktx2 as an option (as ktx2 is already a supper set from which both jpg/png & dds could be derived from); there are use cases where we don't want jpeg/png in addition to ktx2. Currently I3S is keeping jpeg to let client applications catch up supporting ktxt2. Once that happens, next version of I3S will drop all legacy texture formats..

belom88 (approved):

PR #2073, opened 2022-01-28
Add progress to `parseShapefileInBatches`.

nickpeihl:
Based on the work in #1105. This modifies `zipBatchIterators` to return an object with keys `progress` and `data` rather than an array so we can return parsing progress for the `parseShapefileInBatches` function.

This will be useful for returning a progress indicator in a UI when loading large shapefiles.

### DIFF
--- a/modules/shapefile/src/lib/streaming/zip-batch-iterators.ts
+++ b/modules/shapefile/src/lib/streaming/zip-batch-iterators.ts
@@ -1,3 +1,8 @@
+type Batch = {
+  data: number[] | number[][];
+  progress?: {bytesUsed: number; totalBytes: number; rows: number};
+};
+
 /**
  * Zip two iterators together
  *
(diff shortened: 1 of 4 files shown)


modules/shapefile/src/lib/streaming/zip-batch-iterators.ts
```diff
@@ -1,3 +1,8 @@
+type Batch = {
+  data: number[] | number[][];
```
kylebarron, line 2: Why the union? Seems like you'd only need one type?

modules/shapefile/src/lib/streaming/zip-batch-iterators.ts
```diff
@@ -46,17 +51,22 @@ export async function* zipBatchIterators(
...
+  const {data: data1, progress} = batch1;
+  const {data: data2} = batch2;
+  const dataLength: number = Math.min(data1.length, data2.length);
+  if (dataLength === 0) {
     return null;
   }
 
   // Non interleaved arrays
-  const batch: number[][] = [batch1.slice(0, batchLength), batch2.slice(0, batchLength)];
+  const result: any = {
```
kylebarron, line 63: Should this be `result: Batch`?

kylebarron (approved):

nickpeihl:
I'm grooming my backlog and closing stale PRs.

PR #2097, opened 2022-03-01
Geospatial Example: Vancouver data is GeoJSON not GeoPackage

kylebarron:


ibgreen (approved):
Are we planning to add a GeoPackage next? The purpose of this app was really to show that GeoPackage loader is working

kylebarron:
I was thinking a first PR to fix the title and then a second PR after #2098 with a Geopackage example?

PR #2107, opened 2022-03-10
feat(tile-converter): draco writer worker

belom88:


modules/tile-converter/src/i3s-converter/helpers/geometry-converter.d.ts
```diff
@@ -28,7 +28,8 @@ export default function convertB3dmToI3sGeometry(
   attributeStorageInfo: any,
```
ibgreen, line 28: I didn't realize there were still .js and .d.ts files in the loaders.gl code base. Can we move to .ts or is that not possible for tile-converter?

I find it inconvenient to review the .js files below because there are no typings in them.

modules/tile-converter/src/i3s-converter/i3s-converter.ts
```diff
@@ -656,7 +666,8 @@ export default class I3SConverter {
       this.layers0?.attributeStorageInfo,
       this.options.draco,
       this.generateBoundingVolumes,
-      this.geoidHeightModel!
+      this.geoidHeightModel!,
```
ibgreen, line 669: You already know I really don't like exclamation marks ...

modules/core/src/index.ts
```diff
@@ -9,6 +9,7 @@ export {writeFile, writeFileSync} from './lib/fetch/write-file';
 export {setLoaderOptions} from './lib/api/set-loader-options';
 export {registerLoaders} from './lib/api/register-loaders';
 export {selectLoader, selectLoaderSync} from './lib/api/select-loader';
+export {setGlobalOptions, getGlobalOptions} from './lib/loader-utils/option-utils';
```
ibgreen, line 12: Exporting new functions from the core API is kind of a big decision and should be "audited":

- Such functions should have docs and tests...
- There is already a documented public [`setLoaderOptions`](https://loaders.gl/modules/core/docs/api-reference/set-loader-options) does that not work?
- The `getGlobalOptions` seems like a hack, only needed, until we have updated `encode()` functions to handle worker writers? (since the `encode()` function is inside core it could call `getGlobalOptions()` without exporting it... If so maybe we can export `_getLoaderOptions()` with an underscore?
- Should it be called `_getLoaderOptions()` to match the public `setLoaderOptions()`?
-

modules/draco/src/index.ts
```diff
@@ -25,8 +25,10 @@ export const DracoWriterWorker = {
   name: 'Draco compressed geometry writer',
   module: 'draco',
   version: VERSION,
+  worker: true,
   options: {
-    draco: {}
+    draco: {},
+    source: null
```
ibgreen, line 31: I see source here, and workerSource in other places?

modules/tile-converter/src/i3s-converter/helpers/geometry-converter.ts
```diff
@@ -69,7 +69,8 @@ export default async function convertB3dmToI3sGeometry(
   attributeStorageInfo: AttributeStorageInfo[] | undefined,
   draco: boolean,
   generateBoundingVolumes: boolean,
-  geoidHeightModel: Geoid
+  geoidHeightModel: Geoid,
+  workerSource: {[key: string]: string}
```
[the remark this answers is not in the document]
belom88: I load all required workers source code in the initial stage of the conversion. This way source code of a worker is loaded once and is reused for every `WorkerThread` in the pool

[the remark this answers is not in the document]
belom88: Also there is and error in nodejs
`(node:54695) UnhandledPromiseRejectionWarning: TypeError [ERR_WORKER_PATH]: The worker script or module filename must be an absolute path or a relative path starting with './' or '../'. Received "https://unpkg.com/@loaders.gl/draco@3.2.0-alpha.1/dist/draco-writer-nodejs-worker.js"`
It works with source instead of URL so it is loaded out of `worker-utils` scope.

modules/draco/src/index.ts
belom88: It is correspond to the property `source` in `worker-utils`. `workerSource` is  used in `tile-converter` to preliminary load worker sources used by the tile converter.

ibgreen (approved):

PR #2111, opened 2022-03-11
chore(basis): update basis libs

dryabinin-actionengine:
Update basis encoder and transcoder files from https://github.com/BinomialLLC/basis_universal/tree/master/webgl


ibgreen (approved):
Let's add a release note for this.

ibgreen (commented):
I can't see any versioning. I think it would be good to somehow encode which version these libraries are (and certainly mention that in what's new).

dryabinin-actionengine:
@ibgreen - https://github.com/visgl/loaders.gl/pull/2112

PR #2119, opened 2022-03-16
Fix error message truncation in checkResponse

j-esse:
I presume the intention was to use only the first 60 chars - but instead it was removing the first 60 chars.


ibgreen (approved):
Whoops. Thanks!

PR #2131, opened 2022-03-30
chore: refinement-file-converter-documentation

kirillovzah:


dryabinin-actionengine:
1. Fix the name of PR -> chore: refinement-**tile**-converter-documentation.
2. Add more details about what we validate if we turn on validation flag.

PR #2136, opened 2022-04-04
feat(tile-converter): add i3s attributes worker

dryabinin-actionengine:


modules/tile-converter/src/i3s-converter/helpers/i3s-attribute-transformation.ts
```diff
@@ -0,0 +1,42 @@
...
+export type AttributesTransformationOptions = WriterOptions & {
+  attributesTransformations?: AttributesOptions & {
+    useCartesianPostions: false;
+  };
+};
+
+/**
+ * Transform B3DM data to I3S attributes
+ */
+export const I3SAttributeTransformation: Writer = {
```
belom88, line 23: I don't think it is a writer. It should be just worker like in `compression` or `crypto` modules:
"modules/compression/src/compression-worker.ts"

belom88 (changes requested):
We don't need a writer in my opinion. You can call `convertAttributes` directly from the worker.

modules/tile-converter/src/i3s-converter/helpers/i3s-attribute-transformation.ts
dryabinin-actionengine: Done

ibgreen (approved):
Are you able to add any benchmarks to test the performance?

dryabinin-actionengine:
> Are you able to add any benchmarks to test the performance?

Possibly we can do it as separate PR.

belom88 (approved):

PR #2140, opened 2022-04-08
fix(gltf): meshopt compression extension

belom88:
This code is dedicated from https://github.com/visgl/loaders.gl/pull/2035 . 
I checked it with 3DTiles dataset:
image
Textures are in black because of KHR_texture_transform extension that is not supported by @loaders.gl


ibgreen (approved):
We could mention it in what's new, and list it as part of supported extensions in the docs.

PR #2153, opened 2022-04-19
fix(tile-converter): node transformation order

belom88:
Fix the transformation order per spec:
https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-node

Should be `translation * rotation * scale`


ibgreen (approved):
Spec says: `first the scale is applied to the vertices, then the rotation, and then the translation.` I guess we have to apply them in the reverse order.

I think a test case would be good.

belom88:
In the function there is no vertices. We combine the matrix there as per specification.
`vertexVector.transform(nodeMatrix)` is done afterward. I guess it does transformation correctly and multiplies from the correct side, left-hand side.
I checked it on a 3DTiles dataset we have. I don't see the way to test this piece of code. We don't have small datasets for it.

PR #2154, opened 2022-04-20
feat(tile-converter): ktx2 writer worker

belom88:


belom88:
> Any perf measurements? Does it improve conversion speed?

I think so. After write queue implementation I see multiple workers run. I guess we will do compare tests with 3.1.8 after new beta release

PR #2162, opened 2022-04-27
fix(parse-i3s-content): calculate content length

dryabinin-actionengine:


ibgreen (approved):
Does this fix anything or just for readability?

dryabinin-actionengine:
> Does this fix anything or just for readability?

It also helps to enable logic for unloading not needed tiles.
https://github.com/visgl/loaders.gl/blob/master/modules/tiles/src/tileset/tileset-cache.ts#L81-L87

PR #2195, opened 2022-06-12
chore(polyfills): add export for fetchNode

boredland:
I just had the situation where I didn't want to use the whole set of polyfills, but wanted to pass fetchNode to load like this:
```typescript
const batches = await loadInBatches(shapeFilePath, ShapefileLoader, {
    fetch: fetchNode,
});
```

Sadly I think the only way currently is via a file import from `dist`: `@loaders.gl/polyfills/dist/node/fetch/fetch.node`. I'd say we could add an export to make that easier.

### DIFF
--- a/modules/polyfills/src/index.ts
+++ b/modules/polyfills/src/index.ts
@@ -20,7 +20,7 @@ export {BlobPolyfill} from './node/file/blob';
 export {FileReaderPolyfill} from './node/file/file-reader';
 export {FilePolyfill} from './node/file/file';
 export {installFilePolyfills} from './node/file/install-file-polyfills';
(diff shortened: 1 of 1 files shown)


boredland:
> Doesn't fetchNode get installed on globalThis.fetch anyway just by importing loaders.gl/polyfills? Same for other polyfills? So just importing this doesn't really work the way polyfills is current written.

that is true. but as I wrote, I did not want that, because it broke other libraries expecting the vanilla fetch.

> No docs - If we want to export it without adding docs, let's just add an underscore _fetchNode to show that it is not part of the official API.

I think that might be an edge case for me, so adding the underscore!

PR #2209, opened 2022-07-13
fix(draco-builder): handle object metadata properly

dryabinin-actionengine:
`removeNontransferableOptions` function change `Int32Array` to simple `object`.

For example:

const test = {values: new Int32Array([1,2,3,4])};
JSON.stringify(test); // '`{"values":{"0":1,"1":2,"2":3,"3":4}}`'

The fix allows us to make it Int32Array again.


ibgreen:
Typed arrays (at least the underlying array buffers) are transferrable. Should we fix `removeNontransferableOptions` instead of adding a patch to compensate for a bug here?

dryabinin-actionengine:
@ibgreen I'v updated `removeNontransferableOptions` function. Please check.

modules/worker-utils/src/index.ts
```diff
@@ -32,6 +32,7 @@ export {createWorker} from './lib/worker-api/create-worker';
 export {getWorkerURL} from './lib/worker-api/get-worker-url';
 export {validateWorkerVersion} from './lib/worker-api/validate-worker-version';
 export {getTransferList} from './lib/worker-utils/get-transfer-list';
+export {removeNontransferableOptions} from './lib/worker-utils/remove-nontransferable-options';
```
ibgreen, line 35: Why can't we just use `getTransferList`?

dryabinin-actionengine: I checked `getTransferList`  function and for at first sight it should help but it doesn't work as I expected.
First of all it breaks tree structure of options object.
It traverses only one level deep in recursive mode.
It replaced objects with empty array.
And finally 
image
It returns `ArrayBuffer`.

You can see test cases [here](https://github.com/visgl/loaders.gl/blob/master/modules/worker-utils/test/lib/worker-utils/get-transfer-list.spec.js#L16) and there are don't correspond expected result in my case.

ibgreen: OK to speed this up, can you create another function in the same file, e.g. `getTransferListForWriter` and add your implementation there.

At least then I will have both of these implementations in the same place, and we can determine later if we should change the current `getTransferList()`.

It sounds strange but there may be reasons for it working the way it does so I don't want to change it now.

ibgreen (approved):
See my comment about moving your function next to `getTransferList()`

PR #2211, opened 2022-07-14
fix(3d-tiles): add support for absolute content-uris

usefulthink:
Tilesets can use absolute paths or even fully qualified URLs in the
content.uri field. This patch adds support for all of those cases.

How would I best go about testing this? Would it be ok to add a new Tileset to `test/data/Tilesets` specifically to test the url-handling? Or add some unit-tests just for the normalize-functions?

### DIFF
--- a/modules/3d-tiles/src/lib/parsers/parse-3d-tile-header.ts
+++ b/modules/3d-tiles/src/lib/parsers/parse-3d-tile-header.ts
@@ -37,13 +37,27 @@ function getRefine(refine) {
   }
 }
 
+function resolveUri(uri, basePath) {
+  // url scheme per RFC3986
+  const urlSchemeRegex = /^[a-z][0-9a-z+.-]*:/i;
+
+  if (urlSchemeRegex.test(basePath)) {
+    const url = new URL(uri, `${basePath}/`);
+    return decodeURI(url.toString());
+  } else if (uri.startsWith('/')) {
+    return uri;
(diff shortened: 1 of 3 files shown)


modules/3d-tiles/src/lib/parsers/parse-3d-tile-header.ts
```diff
@@ -37,13 +37,25 @@ function getRefine(refine) {
   }
 }
 
+function resolveUri(uri, basePath) {
+  // url scheme per RFC3986
+  const rxUrlScheme = /^[a-z][0-9a-z+.-]*:/i;
```
felixpalmer, line 42: I would call this variable urlSchemeRegex

felixpalmer (changes requested):
I agree it would be good to add some tests to verify some common cases. I would recommend adding a new test file to `modules/3d-tiles/test/lib/parsers/` and testing a set of common combinations by importing the `normalizeXXX` functions and testing those in isolation. I think this is clearer and easier to support than adding a full tileset.json

usefulthink:
Added all the test-cases I could think of for the `normalizeTileData` function. `normalizeImplicitTileHeaders` can't really be tested in isolation since it actually loads the subtree-root and thus requires a working tileset. Those are already tested in `tiles-3d-loader.spec.js`.

felixpalmer (approved):

PR #2232, opened 2022-08-17
fix(worker-utils): check node worker properly

dryabinin-actionengine:


modules/worker-utils/src/lib/worker-farm/worker-thread.ts
```diff
@@ -28,7 +28,10 @@ export default class WorkerThread {
 
   /** Checks if workers are supported on this platform */
   static isSupported(): boolean {
-    return (typeof Worker !== 'undefined' && isBrowser) || typeof NodeWorker !== undefined;
+    return (
+      (typeof Worker !== 'undefined' && isBrowser) ||
+      (typeof NodeWorker !== undefined && !isBrowser)
```
ibgreen, line 33: Is NodeWorker defined in the browser?

ibgreen (approved):

modules/worker-utils/src/lib/worker-farm/worker-thread.ts
dryabinin-actionengine: Yes it is always function. 
image

ibgreen:
> Yes it is always function.

It shouldn't be. It makes me suspect that somehow our node.js code exclusion setup is failing and some kind of polyfill is being included. This looks like a deeper problem. 

What bundler are you using? webpack 4?

dryabinin-actionengine:
> > Yes it is always function.
> 
> It shouldn't be. It makes me suspect that somehow our node.js code exclusion setup is failing and some kind of polyfill is being included. This looks like a deeper problem.
> 
> What bundler are you using? webpack 4?

In i3s example in loaders.gl repo - `"webpack": "^4.39.1",`

PR #2241, opened 2022-09-01
fix(3d-tiles): add options to normalizeTileHeaders

dryabinin-actionengine:
To download implicit tiling subtrees from cesium we need to pass token in options.


ibgreen (approved):
Nit: I would not expect that a function called `normalize` starts loading something. I would just expect it to be a clean data transformation from a format specific data type into a common loaders.gl format. 

Not sure if things could be split into two functions based on this.

dryabinin-actionengine:
> Nit: I would not expect that a function called `normalize` starts loading something. I would just expect it to be a clean data transformation from a format specific data type into a common loaders.gl format.
> 
> Not sure if things could be split into two functions based on this.

Now we have special logic in normalize there we handle and transform implicit tiling to explicit data and there is a function which loads subtrees while such conversion process and it should have loadOptions. Actually I am not sure that it is possible to separate implicit tiling normalization from that place.

PR #2243, opened 2022-09-02
fix(tile-converter): install deps test

belom88:
Prevent `install-dependencies` from rewriting actual workers during tests


ibgreen (approved):
Maybe some comments for the next maintainer?

PR #2244, opened 2022-09-02
fix(polyfills): jpeg parser returns Buffer

belom88:


belom88:
> I wonder if we can improve typing in polyfills to make sure this is rigid all the way?
> 
> If we want the polyfill to return the same format as the image module.

It is output of "get-pixels" library. The library doesn't have types. It returns Uint8Array for PNG and GIF, and Buffer for JPEG. Other then that output of polyfill matches image module.

PR #2247, opened 2022-09-05
chore(i3s): add statistic types

dryabinin-actionengine:


modules/i3s/src/types.ts
```diff
@@ -1005,3 +1005,46 @@ type CameraPosition = {
...
+  /** Representing average or mean value. For example, sum/count. */
+  avg?: number;
+  /** Representing the standard deviation. */
+  stddev?: number;
+  /**	Representing variance. For example, stats.stddev *stats.stddev. */
+  variance?: number;
+  /** Represents the histogram. */
+  histogram?: Histogram;
+  /** An array of most frequently used values within the point cloud scene layer. */
+  mostFrequentValues?: ValueCount;
```
belom88, line 1032: It should be array

belom88 (changes requested):

modules/i3s/src/types.ts
dryabinin-actionengine: FIxed

belom88 (approved):

PR #2255, opened 2022-09-26
chore(i3s): export more types

dryabinin-actionengine:
Also fixed one type.
Spec - https://github.com/Esri/i3s-spec/blob/master/docs/1.8/histogram.cmn.md


ibgreen (approved):
Just consider that every exported type becomes part of the API and changing it becomes a breaking change. It also needs to good docs (at least JSDocs) and ideally we should have minial test cases making sure it the type is defined by the library.

belom88 (approved):

PR #2279, opened 2022-10-24
parquet: slightly simpler arrow api

kylebarron:
I saw https://github.com/visgl/loaders.gl/pull/2276 and thought I should adapt the code to use the upstream `tableFromIPC` and `tableToIPC` while I'm thinking about it

### DIFF
--- a/docs/whats-new.md
+++ b/docs/whats-new.md
@@ -21,7 +21,15 @@
 
 See [roadmap](./roadmap) for more details on planned development.
 
-## v3.2 (In development)
+## v3.3 (In Development)
+
+Release Date: TBD, Q1-2022
+
+**@loaders.gl/arrow**
+
+- Now uses `apache-arrow` v9
+
+## v3.2
 
 Target Release Date: Q1 2022.
 
@@ -58,7 +66,6 @@ Target Release Date: Q1 2022.
 **@loaders.gl/textures**
 
 - Upgrade `basis_universal` libraries to [v1.16.3](https://github.com/BinomialLLC/basis_universal/releases/tag/1.16.3).
-
 ## v3.1
 
 Release Date: Dec 7, 2021.
--- a/modules/arrow/package.json
+++ b/modules/arrow/package.json
@@ -32,12 +32,15 @@
   },
   "scripts": {
     "pre-build": "npm run build-worker && npm run build-bundle",
-    "build-bundle": "esbuild src/bundle.ts --bundle --outfile=dist/dist.min.js",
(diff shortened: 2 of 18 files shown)


ibgreen (approved):
Either way this does require a potentially enormous table to be fully materialized in memory as opposed to streaming it out. 

@trxcllnt may have some thoughts.

kylebarron:
> Either way this does require a potentially enormous table to be fully materialized in memory as opposed to streaming it out.

Yeah. There are two issues. 

- In the current `parquet-wasm` 0.3 API, you have to fetch the entire dataset before you can parse any of it because the API takes in and returns entire `ArrayBuffers`. I added an API since then so that as of 0.4.0-beta.5 you can do the below as an async generator of parquet record batches. (You can't directly stream parquet since the metadata is in the footer at the end of the file, but you can first send a range request for the end metadata, then send range requests for each batch). I haven't gotten around to publishing an 0.4.0 yet because I hit an [upstream webpack bug](https://github.com/rustwasm/wasm-bindgen/issues/3095) and then haven't had time to look into it or revert the `wasm-bindgen` version I use.

    ```js
    import { tableFromIPC } from "apache-arrow";
    // Edit the `parquet-wasm` import as necessary
    import { readRowGroupAsync, readMetadataAsync } from "parquet-wasm";

    const parquetFileMetaData = await readMetadataAsync(url);
    const arrowSchema = parquetFileMetaData.arrowSchema();

    // Read all batches from the file in parallel
    const promises = [];
    for (let i = 0; i < parquetFileMetaData.numRowGroups(); i++) {
      const rowGroupMetaData = parquetFileMetaData.rowGroup(i);
      const rowGroupPromise = wasm.readRowGroupAsync(url, rowGroupMetaData, arrowSchema);
      promises.push(rowGroupPromise);
    }

    // Fetch and parse all of the record batches into a list of Uint8Array IPC buffers
    const recordBatchChunks = await Promise.all(promises);
    // Parse these IPC buffers to Arrow tables
    const chunkTables = recordBatchChunks.map(arrow.tableFromIPC);
    // Create a single output table with multiple chunks
    const table = new arrow.Table(chunkTables);
    ```
- Each time `parquet-wasm` returns data across the rust/wasm -> JS boundary, it makes an extra copy of the table/batch from the Rust Arrow Table representation to an IPC buffer, which it copies into the JS side. It should theoretically be possible to avoid the Rust table -> IPC buffer copy by [implementing the Arrow C Data Interface](https://observablehq.com/@kylebarron/zero-copy-apache-arrow-with-webassembly), but that's a lot of work and a long way off.

PR #2286, opened 2022-11-14
fix(tile-converter): always calculate refinement percentage

dryabinin-actionengine:
We always show statistics about 'ADD' refinement percentage, but calculate it only with validate flag. Now calculation will be performed always.


ibgreen (approved):
Do we still need the option?

dryabinin-actionengine:
> Do we still need the option?

I think we can keep it, we have one more validation depends on this flag.

PR #2334, opened 2023-02-01
chore(parse-basis): add level size field

kirillovzah:


modules/textures/src/lib/parsers/parse-basis.ts
```diff
@@ -254,7 +254,7 @@ function transcodeKTX2Image(ktx2File, levelIndex: number, options): TextureLevel
     compressed,
 
     // Additional fields
-    // Add levelSize field.
+    levelSize: decodedSize,
```
ibgreen, line 257: Is there some precedent for calling this levelSize? Does that already exist on the type or is it a new field? If a new field, should we consider other names? e.g. `byteSize`?

ibgreen (approved):

modules/textures/src/lib/parsers/parse-basis.ts
belom88: This name is already used in dds output. This number is used by luma to calculate texture size. We have an issue in stats when have ktx2 textures.

belom88: image

PR #2385, opened 2023-03-21
fix(tile-converter): node 18 compatibility

belom88:
Related to [#2352](https://github.com/visgl/loaders.gl/pull/2352)


ibgreen (approved):
node14 target is compatible with node18, whereas node12 is not?

belom88:
My idea was NodeJS 12 is out of maintenance so I switched it to NodeJS 14.

PR #2422, opened 2023-04-19
fix(tile-converter): tile-converter dependencies changed and tests enabled

MaximKYZ:
Here is a fix for tile-converter dependences and enabling tile-converter tests that have been disabled in PR https://github.com/visgl/loaders.gl/pull/2410

Here is the end of output for yarn test node:
1..5056
# tests 5056
# pass  5056

# ok

Done in 99.53s.

@ibgreen Could you please review?


dsactionengine:
@ibgreen @belom88 I assume this can be merged?

PR #2432, opened 2023-04-24
fix(wms-example): height of side panel

morphoix:
Before 
Screenshot from 2023-04-24 16-45-45
After
Screenshot from 2023-04-24 16-42-48


jwasilgeo (approved):
Thanks for improving this

PR #2433, opened 2023-04-24
fix(website): i3s debug stats issue

morphoix:
Before
Screenshot from 2023-04-24 18-16-14
After
Screenshot from 2023-04-24 18-16-33


morphoix:
> The fix is good, we need to fix visuals first.
> 
> A separate problem is that many of the loaders.gl examples require too much code, JS / CSS etc. Examples should be small and simple for external users to read and reproduce. Better defaults in the probe.gl stats widget would remove a lot of "clutter" here. Just something to keep in mind as we continue to maintain these examples.

totally agree with you, there is too much of nesting in the examples structure

PR #2458, opened 2023-05-05
Compute memory usage for GLTF tiles

felixpalmer:
Fix for https://github.com/visgl/loaders.gl/issues/2457

### Changes

- Add `getGltfMemoryUsage` for calculating memory usage
- Introduce `gpuMemoryUsageInBytes` getter in `Tile3D`
- Tests


modules/tiles/src/tileset/tile-3d.ts
```diff
@@ -302,6 +301,24 @@ export class Tile3D {
     return this._distanceToCamera;
   }
 
+  /**
+   * Memory usage of tile on GPU
+   */
+  get gpuMemoryUsageInBytes(): number {
```
ibgreen, line 307: I'd drop the getter and just precalculate this in the constructor and add it to a read-only field.

modules/tiles/src/tileset/helpers/memory-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Calculate the GPU memory used by a GLTF tile, for both buffer and texture memory
+ * @param gltf - the gltf content of a GLTF tile
+ * @returns - total memory usage in bytes
+ */
+export function getGltfMemoryUsage(gltf: GLTF) {
+  let {images, bufferViews} = gltf;
+  images = images || [];
+  bufferViews = bufferViews || [];
+  const imageBufferViews = images.map((i) => i.bufferView);
```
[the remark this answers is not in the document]
felixpalmer: That's why I went with 4 bytes per pixel rather than reading the buffer size. As the images I'm working with are jpeg, the buffer size is too small. You're right I didn't consider mipmaps, do you think it is worth adding 33% on top to account for this?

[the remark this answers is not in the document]
ibgreen: Yes that works for JPEG case. However if the tiles are using basis compressed textures (which they should) the GPU/CPU size is 1-to-1  so we would massively overcount texture memory use (probably with a factor of 4 or more). (though note that for compressed textures all mipmap levels are provided rather than autogenerated an must be summed up)

I have a vague recollection that mipmap memory is laid out for access efficiency and not for maximum compactness and that a factor of 1.5 should be used for at least some GPUs. But I guess that is splitting hairs, you may want to define a constant, choose a value and add a comment along these lines.

modules/gltf/src/lib/gltf-utils/gltf-utils.ts
```diff
@@ -77,3 +78,26 @@ export function getAccessorArrayTypeAndLength(accessor, bufferView) {
...
+  let {images, bufferViews} = gltf;
+  images = images || [];
+  bufferViews = bufferViews || [];
+  const imageBufferViews = images.map((i) => i.bufferView);
+  bufferViews = bufferViews.filter((view) => !imageBufferViews.includes(view as any));
+
+  const bufferMemory = bufferViews.reduce((acc, view) => acc + view.byteLength, 0);
+
+  // Assume each pixel of the texture is 4 channel with mimmaps (which add 33%)
+  const pixelCount = images.reduce((acc, image) => {
```
belom88, line 97: This won't work correctly for compressed textures

modules/tiles/src/tileset/tileset-3d.ts
```diff
@@ -871,12 +871,12 @@ export class Tileset3D {
     this.stats.get(TILES_IN_MEMORY).incrementCount();
 
     // Good enough? Just use the raw binary ArrayBuffer's byte length.
-    this.gpuMemoryUsageInBytes += tile.content.byteLength || 0;
+    this.gpuMemoryUsageInBytes += tile.gpuMemoryUsageInBytes || 0;
```
belom88, line 874: You could update tile.byteLength calculation in 3d-tiles module instead of changing it here

belom88 (changes requested):
I would calculate correct `byteLength` in 3d-tiles module. Changing `tiles` module will affect other loaders eg. I3S

modules/gltf/src/lib/gltf-utils/gltf-utils.ts
felixpalmer: Do you know of a way to calculate it for compressed textures? Do we have access to that information from within JS. I would rather overestimate than underestimate

modules/tiles/src/tileset/tile-3d.ts
```diff
@@ -341,6 +342,24 @@ export class Tile3D {
     this._selectedFrame = 0;
   }
 
+  /**
+   * Memory usage of tile on GPU
+   */
+  _getGpuMemoryUsageInBytes(): number {
+    switch (this.type) {
+      case TILE_TYPE.MESH:
```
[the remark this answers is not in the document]
felixpalmer: Sorry, I don't understand this comment, I've maintained the existing behavior for mest&pointcloud which were working fine

modules/tiles/src/tileset/tileset-3d.ts
felixpalmer: Except it isn't the byte length of the tile it is the GPU memory used. Tile3D already has the `gpuMemoryUsageInBytes` field it was just always 0. We even have a comment here asking whether just inspecting the byte length is good enough, and it generally isn't because of the decompressed texture memory being larger than raw bytes

belom88: In that case, could you move `gpuMemoryUsage` calculation to 3d-tiles module? glTF dependency is not good as `tile` module is abstract by design. Feel free to set new property like `content.gpuMemoryUsageInBytes` as `byteLength` is not correct

belom88, line 874:
```suggestion
    this.gpuMemoryUsageInBytes += tile.gpuMemoryUsageInBytes || tile.content.byteLength || 0;
```

modules/tiles/src/tileset/tile-3d.ts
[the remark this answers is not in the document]
belom88: As `gpuMemoryUsage` != `content.byteLength`, returning `content.byteLength` is not correct for `MESH` because there is also can be a texture. I'd suggest returning 0 here to preserve existing logic for `TILE_TYPE.MESH`

modules/tiles/src/tileset/tile-3d.ts
```diff
@@ -341,6 +342,24 @@ export class Tile3D {
...
+      case TILE_TYPE.POINTCLOUD:
+        return this.content.byteLength;
+
+      case TILE_TYPE.SCENEGRAPH:
+        return getMemoryUsageGLTF(this.content.gltf);
```
belom88, line 355: I suggest to move `gpuMemoryUsage` calculation to `3d-tiles` module? `gltf` dependency is not good as `tile` module is abstract by design. Feel free to set new property like `content.gpuMemoryUsageInBytes` as `content.byteLength` is not correct

modules/tiles/src/tileset/tile-3d.ts
```diff
@@ -341,6 +342,24 @@ export class Tile3D {
...
   }
 
+  /**
+   * Memory usage of tile on GPU
+   */
+  _getGpuMemoryUsageInBytes(): number {
+    switch (this.type) {
+      case TILE_TYPE.MESH:
+      case TILE_TYPE.POINTCLOUD:
+        return this.content.byteLength;
```
belom88, line 352: Probably, it would be nice to set `content.gpuMemoryUsageInBytes = content.byteLength` for `POINTCLOUD` in `3d-tiles`.

modules/tiles/src/tileset/tile-3d.ts
```diff
@@ -341,6 +342,24 @@ export class Tile3D {
...
+
+      case TILE_TYPE.SCENEGRAPH:
+        return getMemoryUsageGLTF(this.content.gltf);
+
+      case TILE_TYPE.EMPTY:
```
belom88, line 357: ditto as for POINTCLOUD

belom88 (changes requested):

modules/tiles/src/tileset/tile-3d.ts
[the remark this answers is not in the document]
felixpalmer: I've left if as returning `content.byteLength` as it is more accurate than 0. If there is a reason you think it will break something I can special case it to 0

modules/gltf/src/lib/gltf-utils/gltf-utils.ts
felixpalmer: I can take a look at this in a followup PR

modules/gltf/src/index.ts
```diff
@@ -42,6 +42,7 @@ export {GLBWriter} from './glb-writer';
 // glTF Data Access Helper Class
 export {default as GLTFScenegraph} from './lib/api/gltf-scenegraph';
 export {postProcessGLTF} from './lib/api/post-process-gltf';
+export {getMemoryUsageGLTF} from './lib/gltf-utils/gltf-utils';
```
ibgreen, line 45: Nit: This should probably be an underscore export to mark it as internal / not audited.

modules/gltf/src/lib/gltf-utils/gltf-utils.ts
```diff
@@ -77,3 +78,27 @@ export function getAccessorArrayTypeAndLength(accessor, bufferView) {
   assert(byteLength >= 0 && byteLength <= bufferView.byteLength);
   return {ArrayType, length, byteLength};
 }
+
+/**
+ * Calculate the GPU memory used by a GLTF tile, for both buffer and texture memory
+ * @param gltf - the gltf content of a GLTF tile
+ * @returns - total memory usage in bytes
+ */
+export function getMemoryUsageGLTF(gltf: GLTF) {
```
ibgreen, line 87:
```suggestion
export function getMemoryUsageGLTF(gltf: GLTF): number {
```

modules/gltf/src/lib/gltf-utils/gltf-utils.ts
```diff
@@ -77,3 +78,27 @@ export function getAccessorArrayTypeAndLength(accessor, bufferView) {
...
+  images = images || [];
+  bufferViews = bufferViews || [];
+  const imageBufferViews = images.map((i) => i.bufferView);
+  bufferViews = bufferViews.filter((view) => !imageBufferViews.includes(view as any));
+
+  const bufferMemory = bufferViews.reduce((acc, view) => acc + view.byteLength, 0);
+
+  // Assume each pixel of the texture is 4 channel with mimmaps (which add 33%)
+  // TODO correctly handle compressed textures
+  const pixelCount = images.reduce((acc, image) => {
```
ibgreen, line 98: For later: We could create a similar `getMemoryUsageForTexture` function in the `textures` module that would calculate the size of a texture (which includes images).

We could gradually improve it to handle compressed textures.

belom88 (changes requested):

PR #2484, opened 2023-05-17
fix(hero example): change dataset url

morphoix:
Screenshot from 2023-05-17 13-25-49


ibgreen (approved):
Thanks. Does this example load faster than the other one?

ibgreen (approved):
Thanks. Does this example load faster than the other one?

morphoix:
> Thanks. Does this example load faster than the other one?

it seems like this. At least textures download faster

PR #2498, opened 2023-05-22
WIP: Tile3D getCoordinatesByIndex

radoslavirha:
Hi,

regarding to my previous discussion https://github.com/visgl/loaders.gl/discussions/2425#discussion-5114144 I prepared something which I can improve with your guidance.

My final target is to create cursor similar to [this](https://codesandbox.io/s/dof-forked-v8jl2b?file=/src/App.js:1263-1275).

On meshes it should be simple, maybe another method can return orientation of mesh under cursor + cursor coords and I can render oriented circle under cursor.

On point clouds it's more difficult I think, so I started with picking multiple points around cursor and I want to create plane from those (3?) points, calculate orientation,... So this is the use case where it's useful for me. But it could be usable generally too.

Thanks

### DIFF
--- a/modules/tiles/src/tileset/tile-3d.ts
+++ b/modules/tiles/src/tileset/tile-3d.ts
@@ -5,6 +5,7 @@
 
 import {Vector3, Matrix4} from '@math.gl/core';
 import {CullingVolume} from '@math.gl/culling';
+import {addMetersToLngLat} from '@math.gl/web-mercator';
 
 import {load} from '@loaders.gl/core';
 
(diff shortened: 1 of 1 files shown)


ibgreen:
Thanks for putting this together, and apologies that it did not receive a timely review.

The use case is valid, but I don't think this implementation can be merged as written:

- `attributes.positions` contains tile-local positions. They need to be transformed through the content's `cartographicModelMatrix` or `cartesianModelMatrix` before conversion to WGS84. Passing the raw values directly to `addMetersToLngLat` gives incorrect results for transformed tiles, `RTC_CENTER`, recentered content, and quantized data.
- Quantized point clouds normally expose positions as an accessor object (`{value, size, normalized, ...}`), so indexing `attributes.positions` directly produces undefined/NaN coordinates.
- Invalid, fractional, and out-of-range indices are not validated.
- A public API addition would also need TSDoc, documentation, and coverage for ordinary, transformed, RTC-centered, and quantized point clouds.

The branch is now several years behind master, has conflicts, and the `Tile3D` implementation has moved. Given that, I'm closing this PR rather than asking for a substantial rebase and redesign.

If this capability is revisited, a fresh PR should probably start with a generic helper that transforms a supplied tile-local position to Cartesian or cartographic coordinates, with an optional point-cloud-specific indexed accessor layered on top. For exact WGS84 output, applying `cartesianModelMatrix` and then `Ellipsoid.WGS84.cartesianToCartographic` would be a more robust path.

PR #2532, opened 2023-06-30
fix(tile-converter): install deps on windows

belom88:


dsactionengine (approved):
process.platform === 'win32' ? 'npm.cmd' : 'npm'
works as expected - installs on windows

PR #2547, opened 2023-07-12
chore(tile-converter): Support for SLPKs larger than 2 Gb

dariaterekhovaae:


belom88 (changes requested):

ibgreen (approved):

modules/i3s/test/zip-utils/data-view-provider.spec.js
```diff
@@ -0,0 +1,34 @@
+import test from 'tape-promise/tape';
```
belom88, line 1: FYI: It is possible to make *.ts tests

modules/i3s/src/lib/parsers/parse-zip/signature.ts
```diff
@@ -0,0 +1,2 @@
+/** Description of zip signature type */
+export type ZipSignature = [number, number, number, number];
```
belom88, line 2: Nit. The type is so simple. The better is just declare a variable with this tuple:
```
const zipSignature: [number, number, number, number] = [1,2,3,4];
```
The separate file is too much for it.

belom88 (approved):

PR #2560, opened 2023-07-19
fix(tile-converter): CesiumION tileset URL

belom88:


modules/tile-converter/src/i3s-converter/i3s-converter.ts
```diff
@@ -240,10 +240,14 @@ export default class I3SConverter {
 
     try {
       const preloadOptions = await this._fetchPreloadOptions();
+      let tilesetUrl = inputUrl;
```
ibgreen, line 243: The code is a bit difficult to follow. I think it is partly because the existing helper function is called `_fetchPreloadOptions` but this update makes it clear that the main thing it returns a URL, and beyond that only the headers of the options are used.

`_preloadURLAndHeaders()` ?

ibgreen (approved):
I guess it is hard to write a test for this case.

modules/tile-converter/src/i3s-converter/i3s-converter.ts
belom88: The response of prefetch:
image
There are other properties that we don't need

PR #2572, opened 2023-08-03
fix(tile-converter): fix loading buffers in preprocess-3d-tiles

mspivak-actionengine:


ibgreen (approved):

belom88 (changes requested):
This change doesn't fix the problem in `gltf`. The problem is that `KHR_texture_transform` extension fails if buffers are not loaded.
I suggest making a change in https://github.com/visgl/loaders.gl/blob/f51905b1385c8ba51c238adc70868fed997474fa/modules/gltf/src/lib/extensions/KHR_texture_transform.ts#L64
```
if (!hasExtension || !options.gltf?.loadBuffers) {
```
A test case in `modules/gltf/test/lib/extensions/KHR_texture_transform.spec.js` will be appreciated.

modules/gltf/src/lib/api/post-process-gltf.ts
```diff
@@ -135,6 +135,12 @@ class GLTFPostProcessor {
     this.images = images;
     this.jsonUnprocessed = json;
 
+    if (!options?.loadBuffers) {
+      // @ts-expect-error
+      const jsonPost: GLTFPostprocessed = {...json};
```
belom88, line 140: It doesn't return postprocessed gltf that is anoying for a user. I don't agree with this change

modules/gltf/test/lib/extensions/KHR_texture_transform.spec.js
```diff
@@ -25,3 +25,22 @@ test('GLTFLoader#KHR_texture_transform', async (t) => {
   );
   t.end();
 });
+
+test('GLTFLoader#KHR_texture_transform with no buffers loaded', async (t) => {
+  const response = await fetchFile(GLTF_BINARY_URL);
+  const data = await response.arrayBuffer();
+  /**
+   * @type {import("@loaders.gl/gltf").ParseGLTFOptions}
+   */
```
belom88, line 34: Convert the file to TypeScript probably

modules/gltf/test/lib/extensions/KHR_texture_transform.spec.js
```diff
@@ -25,3 +25,22 @@ test('GLTFLoader#KHR_texture_transform', async (t) => {
...
+  const loaderOptions = {loadBuffers: false};
+  const gltfWithBuffers = await parse(data, GLTFLoader, {gltf: loaderOptions});
+  const gltf = postProcessGLTF(gltfWithBuffers, loaderOptions);
```
belom88, line 37: We don't do postprocess gltf without buffer usually

belom88 (approved):

PR #2574, opened 2023-08-03
chore(tile-converter): 3dtiles - exclude Tile3D and Tileset3D

belom88:
Final step for #2517


ibgreen (approved):
Looks like a solid typing pass. Much appreciated!

PR #2585, opened 2023-08-14
fix(i3s): add  to tileset data

belom88:


MaximKYZ (approved):
Looks and works fine

ibgreen (changes requested):
There is a problem here:

Returned data from loaders should preferentially be a pure JavaScript object, that is easy to serialize etc.

But the loader object is not a pure JavaScript, it has functions etc. 

Functions and methods should be in the helper classes such as Tileset3D. Perhaps we can add a parameter there?

belom88:
> There is a problem here:
> 
> Returned data from loaders should preferentially be a pure JavaScript object, that is easy to serialize etc.
> 
> But the loader object is not a pure JavaScript, it has functions etc.
> 
> Functions and methods should be in the helper classes such as Tileset3D. Perhaps we can add a parameter there?

This is a regression. Similar `loader` property is returned by `3d-tiles` loader and @deck.gl expects this `loader` property. We could plan some refactoring on future but now it is broken without this change

PR #2596, opened 2023-08-23
Fix browser exclude

Pessimistress:
For https://github.com/visgl/deck.gl/issues/8077


modules/loader-utils/src/lib/node/fs.ts
```diff
@@ -4,74 +4,37 @@ import {toArrayBuffer} from './buffer';
...
 /** Wrapper for Node.js fs method */
-export let open;
+export const open: any = fs.open;
 /** Wrapper for Node.js fs method */
-export let close: (fd: number) => Promise<void>;
+export const close = (fd: number) =>
+  new Promise<void>((resolve, reject) => fs.close(fd, (err) => (err ? reject(err) : resolve())));
 /** Wrapper for Node.js fs method */
-export let read;
+export const read: any = fs.read;
```
Pessimistress, line 30: Forcing type to `any` to disable TS errors.

I do not think these functions are being used correctly, but I don't know enough about the system to fix them.

ibgreen (commented):
Thanks. I suggest fixing these on 3.4-release branch instead. If things work well I can make a pass and land this on master.

PR #2629, opened 2023-09-07
chore(gis): Add typescript types

ibgreen:


kylebarron (commented):
From what I remember this looks good

PR #2640, opened 2023-09-11
fix(tile-converter): geometry attributes reordering performance

belom88:


ibgreen (approved):

modules/tile-converter/src/i3s-converter/helpers/geometry-attributes.ts
```diff
@@ -195,35 +205,35 @@ function getSliceAttributeCount(
...
-      existingObject.colors = concatenateTypedArrays(existingObject.colors, currentObject.colors);
-      existingObject.texCoords = concatenateTypedArrays(
-        existingObject.texCoords,
-        currentObject.texCoords
-      );
-      existingObject.uvRegions = concatenateTypedArrays(
-        existingObject.uvRegions,
-        currentObject.uvRegions
-      );
+      existingObject.length += data.positions.length / 3;
```
mspivak-actionengine, line 216: Maybe VALUES_PER_VERTEX  instead of 3 ?

modules/tile-converter/src/i3s-converter/helpers/geometry-attributes.ts
```diff
@@ -267,8 +265,43 @@ function groupAttributesAndRangesByFeatureId(
     objIndex += 1;
   }
 
-  range.push(positions.length / POSITIONS_AND_NORMALS_PER_TRIANGLE - 1);
+  const attributes = concatenateAttributes(unifiedObjects);
+
```
mspivak-actionengine, line 269: Are triangles only supported? Would it be necessary to check the type of objects?

mspivak-actionengine (approved):

modules/tile-converter/src/i3s-converter/helpers/geometry-attributes.ts
belom88: Removed `length` property. It is not needed anymore

PR #2648, opened 2023-09-14
docs(release-notes): Loaders 4.0 upcoming release notes

dsactionengine:
Additions from ActionEngine team into Loaders.gl roadmap and/or whats-new pages for the upcoming 4.0 release of Loaders.gl


ibgreen (approved):
Let's put anything that has made it in, into `whats-new.md` instead or `roadmap`.

We are polishing our release notes there, it is OK if it is not completely correct up until we release.

PR #2671, opened 2023-10-02
fix: getting tile url with empty query params

maxkuznetsov-actionengine:
We don't need sign '?' at the end of the url if there are no query parameters

### DIFF
--- a/modules/tiles/src/tileset/tileset-3d.ts
+++ b/modules/tiles/src/tileset/tileset-3d.ts
@@ -348,9 +348,12 @@ export class Tileset3D {
     if (isDataUrl) {
       return tilePath;
     }
-    return `${tilePath}${tilePath.includes('?') ? '&' : this.queryParams.length ? '?' : ''}${
-      this.queryParams
-    }`;
+
+    let tileUrl = tilePath;
+    if (this.queryParams.length) {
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Add test case?

modules/tiles/src/tileset/tileset-3d.ts
```diff
@@ -348,7 +348,9 @@ export class Tileset3D {
     if (isDataUrl) {
       return tilePath;
     }
-    return `${tilePath}${tilePath.includes('?') ? '&' : '?'}${this.queryParams}`;
+    return `${tilePath}${tilePath.includes('?') ? '&' : this.queryParams.length ? '?' : ''}${
```
felixpalmer, line 351: I think this code is too dense, could you split across multiple lines? You're handling the `queryParams = ''` case, but it still isn't perfect as `&` is needlessly appended if the tilePath contains `?`

felixpalmer (changes requested):

modules/tiles/src/tileset/tileset-3d.ts
maxkuznetsov-actionengine: Thanks, agree. Please review updated version.

felixpalmer (approved):

belom88 (approved):

PR #2680, opened 2023-10-09
fix: WMS 1.3.0 compatibility on GetFeatureInfo

mortenolsenzn:
`getMapURL` uses the `_getWMS130Parameters` to improve compatibility between WMS 1.3.0 and 1.1.0, but `getFeatureInfoUrl` does not.

This means, for instance if SRS is set to `EPSG:3857` it will be handled incorrectly by `getFeatureInfoUrl` as it would use the default CRS of `EPSG:4326`, causing bbox flipping

### DIFF
--- a/modules/wms/src/services/ogc/wms-service.ts
+++ b/modules/wms/src/services/ogc/wms-service.ts
@@ -363,6 +363,7 @@ export class WMSSource extends ImageSource<WMSSourceProps> {
     wmsParameters: WMSGetFeatureInfoParameters,
     vendorParameters?: Record<string, unknown>
   ): string {
+    wmsParameters = this._getWMS130Parameters(wmsParameters);
     const options: Required<WMSGetFeatureInfoParameters> = {
       version: this.wmsParameters.version,
       // query_layers: [],


modules/wms/src/services/ogc/wms-service.ts
```diff
@@ -363,6 +363,7 @@ export class WMSSource extends ImageSource<WMSSourceProps> {
     wmsParameters: WMSGetFeatureInfoParameters,
     vendorParameters?: Record<string, unknown>
   ): string {
+    wmsParameters = this._getWMS130Parameters(wmsParameters);
```
JensRoland, line 366: What is this scoping magic? does `wmsParameters = ` assignment assign to `this.wmsParameters` ?

No JavaScript, that's a bad JavaScript!
image

ibgreen (approved):
Thanks for the PR.

I *think* this will work, though it could be possible that some parameter for getFeatureInfo needs different treatment. But let's land this and we'll see how it works.

It would be nice to add at least one test case, I think there are a bunch of test cases for `getMapURL`, you should be able to add at least one test case for `getFeatureInfoURL`.

There are a couple of additional more rarely used WMS requests, I don't know if they need similar treatment. But should be safe to leave them out for now.

mortenolsenzn:
Yes test cases would be nice, and would like to add them, but I have not heard about WMS or CRS until yesterday, so I might not be the best to assess what would be the correct behaviour in different scenarios.

The reason I stumbled into this is because the experimental `deck.gl` [WMS layer](https://deck.gl/docs/api-reference/geo-layers/wms-layer) is written for the 1.1.0 spec (or at least use the SRS property instead of CRS).

Another issue, that might be worth fixing, but which I left our of the PR because I don't know the correct solution is the change between 1.1.0 and 1.3.0, where `X` and `Y` was changed to I and J and represent the same order as specified in the given CRS.

So if it should correctly preserve compatibility with 1.1.0 request, I guess it should convert `X` to `I` and `Y` to `J` for any lat-lon CRS and `X` to `J` and `Y` to `I` for any lon-lat CRS

At least right now the Deck.GL layer would fail on fetching feature info, since it would set the version to 1.3.0 but send the map points as X and Y

But again, this is the first time I work with the format, so I will not pretend that I have good grasp of the specification.

ibgreen:
> Yes test cases would be nice, and would like to add them, but I have not heard about WMS or CRS until yesterday, so I might not be the best to assess what would be the correct behaviour in different scenarios.

There are already a number of test cases that just make sure that the functions generate the intended string, so that we catch any regressions. Adding one more test case there should not be hard.

> 
> The reason I stumbled into this is because the experimental `deck.gl` [WMS layer](https://deck.gl/docs/api-reference/geo-layers/wms-layer) is written for the 1.1.0 spec (or at least use the SRS property instead of CRS).
> 
Yes, we need to upgrade deck, feel free to make a PR on deck.gl

> Another issue, that might be worth fixing, but which I left our of the PR because I don't know the correct solution is the change between 1.1.0 and 1.3.0, where `X` and `Y` was changed to I and J and represent the same order as specified in the given CRS.
> 
> So if it should correctly preserve compatibility with 1.1.0 request, I guess it should convert `X` to `I` and `Y` to `J` for any lat-lon CRS and `X` to `J` and `Y` to `I` for any lon-lat CRS

You lost me here. Do you have a link to a doc or spec that introduces new parameters `i` and `j`?

mortenolsenzn:
Great, I will add the test case tomorrow, and make a PR for upgrading deck.gl's WMS layer to support 1.3.0

As for the change from X and Y to I and J i found it in [OpenGIS' WMS 1.3.0 implementation specification](https://raw.githubusercontent.com/visgl/deck.gl-data/master/specifications/wms/06-042_OpenGIS_Web_Map_Service_WMS_Implementation_Specification.pdf) in section 7.4.2 or more precise 7.4.3.7

> The mandatory I and J request parameters are integers that indicate a point of interest on the map that was
produced by the embedded GetMap request (the “map request part” described in 7.4.3.3). The point (I,J) is a
point in the (i,j) space defined by the Map CS (see 6.7.2). 

Compaired with section 7.3.3.8 of the [version 1.1.0 implementation specification](https://portal.ogc.org/files/?artifact_id=1081&format=pdf)

> The required X and Y parameters indicate a point of interest on the map. X and Y
identify a single point within the borders of the WIDTH and HEIGHT parameters of the
embedded GetMap request. The origin is set to (0,0) centered in the pixel at the upper left
corner; X increases to the right and Y increases downward.

PR #2690, opened 2023-10-11
fix(tiles): cartographicToCartesan syntax

belom88:
replace
```
const x = Ellipsoid.WGS84.cartographicToCartesian(y, new Vector3());
```
with
```
const x = new Vector3();
Ellipsoid.WGS84.cartographicToCartesian(y, x)
```
Existing syntax fails in JEST tests environment in "I3S Explorer". It uses es5 transpiled code and `cartographicToCartesian` returns `undefined`. The issue occured from NodeJS 16. NodeJS 14 is ok.


ibgreen (approved):
I am not too excited about making such a change when we don't fully understand what is causing it, but since the changes are quite small and unblocks work...

PR #2708, opened 2023-10-18
fix(tile-converter): enable tests

belom88:


ibgreen (approved):
Looks great, but since I just landed a PR that bumps tests to Node 18, please rebase and recheck that this passes CI before landing.

PR #2713, opened 2023-10-20
fix(i3s): I3SContentLoader regression

belom88:


ibgreen (approved):
Generally we should use context.fetch, context.parse, context.load inside loaders instead of importing core. (ideally loader modules shouldn't even import core).

Perhaps core just need to be modified to set context.fetch to fetchFile?

PR #2734, opened 2023-10-25
fix(website): restore I3S examples

belom88:


ibgreen (approved):
Are we removing the i3s-debug example permanently in favor of i3s-explorer? 

Should we also consider deleting the source code for that example? Perhaps a topic for our next sync meeting?

PR #2765, opened 2023-11-02
chore(Arrow): add test cases for geoarrow to binary geometries

lixun910:
- add test cases for `getBinaryGeometriesFromArrow()`
- Fix lint


ibgreen (approved):

ibgreen (approved):
I wonder how hard it would be to create a version that took an `AsyncIterator<arrow.RecordBatch>` and converted batch by batch?

PR #2776, opened 2023-11-06
fix(examples): run 'geospatial' locally

belom88:
It still fails because of recently enabled `ParquetWasmLoader`. But now I can comment it to debug the JS loader.


ibgreen (approved):
Thanks for doing this. I will need to fix this node dependency but for now this helps.

PR #2808, opened 2023-11-27
GLTF: inline emod function to remove loaders.gl/math dependency

felixpalmer:
Without this, importing `@loaders.gl/gltf` fails (e.g. in luma.gl) due to this import: https://github.com/visgl/loaders.gl/blob/master/modules/gltf/src/lib/extensions/utils/3d-tiles-utils.ts#L17


ibgreen (approved):
Nit: While I really appreciate the effort to modularize, I feel that adding a dependency just for a one liner mod function is overkill. While tree shaking should take care of removing all the unused functions from production bundles, every dependency does make the library a little heavier to use.

In this case I would just copy the emod function over to the importing file and remove the dependency.

PR #2836, opened 2023-12-11
feat(GeoArrow): add optional delay between batches in parseArrowInBatches

lixun910:
Description: add an optional delay between batches in parseArrowInBatches, so the incremental loading and rendering based on batches can be smooth e.g. in https://github.com/keplergl/kepler.gl/pull/2474


ibgreen (approved):

ibgreen (commented):
https://github.com/visgl/loaders.gl/blob/9b357878d28a62d0f8024577426f9aca8a04091a/modules/schema/src/lib/table/batches/table-batch-builder.ts#L142

lixun910:
Add `options?: ArrowLoaderOptions` in `parseArrowInBatches`. Then we can control the delay between batches by passing down the `batchDebounceMs` parameter from e.g. kepler

PR #2845, opened 2023-12-18
chore: Add missing icon for I3S BSL example

maxkuznetsov-actionengine:
Add missing icon, the same as we have for I3S example for now


ibgreen (approved):
This thumbnail doesn't really show anything BSL specific. it is just another SF screenshot. maybe we can improve it later?

maxkuznetsov-actionengine:
Correct, it's just to add missing icon. We have to change it to something else in the future.

PR #2847, opened 2023-12-20
Update the empty traversal algorithm to match CesiumJS, possible fix for low-res flickering

Avnerus:
Hi!
After a year of absence I'm back doing some viz work. 
I'm now updating [nytimes/three-loader-3dtiles](https://github.com/nytimes/three-loader-3dtiles/tree/feature/loadersgl-4) to loaders.gl-4 and support for Google's 3D Tiles (*feature/loadersgl-4* branch).
I noticed one issue where low-res tiles flicker above the high-res tiles during traversals, possible related to [this issue](https://github.com/visgl/loaders.gl/issues/2570). It seemed to have come from some glitch in the "empty" traversal mechanism. I updated the algorithm to match the [latest CesiumJS one](https://github.com/CesiumGS/cesium/blob/66dba4b2e6bef81e5662499921c2eacd35a90dee/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L251C25-L251C26) and at least on my tests it seems to fix the issue and does not hinder performance. Could you also run some tests on your side, deck.gl and such?
And if this is good to merge, could you also merge it as a fix to the stable loaders.gl v4?
Thank you!
/Avner

### DIFF
--- a/modules/tiles/src/tileset/tileset-traverser.ts
+++ b/modules/tiles/src/tileset/tileset-traverser.ts
@@ -335,38 +335,35 @@ export class TilesetTraverser {
   executeEmptyTraversal(root: Tile3D, frameState: FrameState): boolean {
     let allDescendantsLoaded = true;
     const stack = this._emptyTraversalStack;
-
     stack.push(root);
 
-    while (stack.length > 0 && allDescendantsLoaded) {
+    while (stack.length > 0) {
       const tile = stack.pop();
+
+      const traverse = !tile.hasRenderContent && this.canTraverse(tile, frameState, false, false);
+      const emptyLeaf = !tile.hasRenderableContent && tile.children.length === 0;
+
+      // Traversal stops but the tile does not have content yet
+      // There will be holes if the parent tries to refine to its children, so don't refine
+      // One exception: a parent may refine even if one of its descendants is an empty leaf
+      if (!traverse && !tile.contentAvailable && !emptyLeaf) {
+        allDescendantsLoaded = false;
+      }
 
       this.updateTile(tile, frameState);
 
       if (!tile.isVisibleAndInRequestVolume) {
-        // Load tiles that aren't visible since they are still needed for the parent to refine
         this.loadTile(tile, frameState);
+        this.touchTile(tile, frameState);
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
@Avnerus Welcome back!

It is possible to add a test case or two? Traversal tests are a weak area in loaders.gl so maybe it is too much to ask but I think we need to start somewhere.

@Kaapp and @belom88  if not a detailed review, would appreciate at least rubberstamps showing that you have scanned the code.

If this doesn't make it into a 4.0 patch release, we are pretty close to releasing 4.1. Let's decide when it lands.

Avnerus:
Thanks @ibgreen. 
I pushed another commit, there was a wrong property name taken from CesiumJS.
I looked some more into the code and looks like this was the problematic line:
```
      while (stack.length > 0 && allDescendantsLoaded) {
```
And the problematic `allDescendantsLoaded` condition was added by yours truly [here](https://github.com/visgl/loaders.gl/pull/1183), perhaps copying older code of CesiumJS, or perhaps it was my idea - I'm not sure. But looking toward a test case - there could be a parent tile where one child answers the condition of `!allDescendantsLoaded`, that breaks the traversal and probably prevents the loading of sibling tiles that do need to be refined.

Kaapp (approved):
nice find! looks good to me 🎉 

looking at the linked issue i agree that this was likely also the culprit

felixpalmer:
@Avnerus really great to see this, this issue has been a pain for a while. Will integrate into deck.gl soon

Avnerus:
@ibgreen, do you think this could be a patch for v4.0? I prefer to wait for this merge before updating [three-loader-3dtiles](https://github.com/nytimes/three-loader-3dtiles) to support Google's 3D Tiles.

felixpalmer:
@Avnerus it is published in the [latest 4.1 alpha](https://github.com/visgl/loaders.gl/blob/master/CHANGELOG.md)

Avnerus:
@felixpalmer Oh I didn't realize that the alpha builds are also published to npm. Thanks! I'll give it a go.

PR #2849, opened 2023-12-29
feat(tile-converter): --add-hash mode

dariaterekhova-actionengine:


modules/tile-converter/src/converter-cli.ts
```diff
@@ -92,6 +100,31 @@ async function main() {
...
 
+  if (options.addHash) {
+    if (!options.tileset) {
+      return;
+    }
+    let finalPath = options.tileset;
+    if (!options.output || options.output === 'data') {
+      const nameWithoutExt = options.tileset.substring(0, options.tileset.length - 5);
+      const response = prompt(
+        `Would you like to change the current file? If no, a new file ${nameWithoutExt}-hash.slpk with hash file inside will be created (Y/N): `
```
belom88, line 111: Just a suggestion:
```suggestion
        `Would you like to modify the current file? If no, a new file ${nameWithoutExt}-hash.slpk with hash file inside will be created (Y/N): `
```

belom88 (approved):

PR #2856, opened 2024-01-13
fix(loader-utils): concatenateArrayBuffers bottleneck fix

dariaterekhova-actionengine:
I've got the `RangeError: Maximum call stack size exceeded` error for huge slpk creation. So I investigated and found out that we have a bottleneck on the current implementation of concatenateArrayBuffers function in amount of arguments. I haven't found any info about nodejs, but I've found the following article: https://stackoverflow.com/a/22747272 and it looks like node has the same limitation of max 65 535 arguments per function and throw the same error since it use the same V8 engine to interpret JS code. But number of files even in 5gb datasets can reach 800 000, so node can't handle it and throws an error.

To solve this problem I've created another function receiving `ArrayBuffer[]` instead of multiple arguments

### DIFF
--- a/modules/loader-utils/src/index.ts
+++ b/modules/loader-utils/src/index.ts
@@ -64,6 +64,7 @@ export {parseJSON} from './lib/parser-utils/parse-json';
 export {
   sliceArrayBuffer,
   concatenateArrayBuffers,
+  concatenateArrayBuffersFromArray,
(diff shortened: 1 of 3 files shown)


ibgreen (approved):
Good catch, classic JS pitfall. 

Would be good to convert all calls to `concatenateArrayBuffers`.

PR #2871, opened 2024-01-30
chore(zip): zip composition feature integration

dariaterekhova-actionengine:
Integration of own algorithm to compose slpk in order to get rid of zip/7zip console archivers usage


ibgreen (approved):
Looks good to me, but please run `yarn test browser` to check that all tests work

PR #2873, opened 2024-01-30
fix(tile-converter): add default refine value to the root

mspivak-actionengine:


modules/tile-converter/src/3d-tiles-converter/3d-tiles-converter.ts
```diff
@@ -121,6 +121,9 @@ export default class Tiles3DConverter {
     await this._addChildren(rootNode, rootTile, 1);
 
     const tileset = transform({root: rootTile}, tilesetTemplate());
+    if (!tileset.root.refine) {
```
belom88, line 124: Set it in the tile initialization https://github.com/visgl/loaders.gl/blob/eec7089c91b2da0a42756aa149469967676c4087/modules/tile-converter/src/3d-tiles-converter/3d-tiles-converter.ts#L113 . There is no a case when it can be "ADD"

belom88 (changes requested):

modules/tile-converter/src/3d-tiles-converter/3d-tiles-converter.ts
```diff
@@ -121,6 +121,9 @@ export default class Tiles3DConverter {
...
+      tileset.root.refine = 'REPLACE';
```
belom88, line 125: It would be good to validate this value in the test cases

belom88 (changes requested):

PR #2875, opened 2024-01-31
fix(tile-converter): skip showing progress with no analyze

mspivak-actionengine:


modules/tile-converter/src/i3s-converter/helpers/progress.ts
```diff
@@ -119,7 +119,7 @@ export class Progress {
    * @returns Number of milliseconds remaining
    */
   getTimeRemaining(): {timeRemaining: number; trust: boolean} | null {
-    if (!this._stepsDone || !this.startTime) {
+    if (!this._stepsTotal || !this._stepsDone || !this.startTime) {
```
belom88, line 122: Can we add a test for this corner case?

modules/tile-converter/src/lib/utils/statistic-utills.ts
```diff
@@ -44,7 +44,7 @@ function timeConverterFromSecondsAndMilliseconds(timeInSeconds: number, millisec
   }
 
   if (!result) {
-    result += `${milliseconds}ms`;
+    result += `${Math.floor(milliseconds)}ms`;
```
belom88, line 47: Can we add a test for this change?

belom88 (changes requested):

modules/tile-converter/src/lib/utils/statistic-utills.ts
mspivak-actionengine: Done

belom88 (approved):

PR #2887, opened 2024-02-19
chore(i3s examples): add i3s colorization by attributes

maxkuznetsov-actionengine:


belom88 (changes requested):
This is too large for an example.
I'd just have a dropdown with attributes to colorize. We don't need to have histogram, set of icons, picking and so on.
Could we make colorization and i3s removal in different PRs?

dsavinov-actionengine:
> This is too large for an example. I'd just have a dropdown with attributes to colorize. We don't need to have histogram, set of icons, picking and so on. Could we make colorization and i3s removal in different PRs?

1. For the UI, would you both @maxkuznetsov-actionengine @belom88 agree with this?
image

2.  Splitting in 2 different PRs - ok.

belom88:
> > This is too large for an example. I'd just have a dropdown with attributes to colorize. We don't need to have histogram, set of icons, picking and so on. Could we make colorization and i3s removal in different PRs?
> 
> 1. For the UI, would you both @maxkuznetsov-actionengine @belom88 agree with this?
>    image
> 
> For Colorize Mode: None/Replace/Multiply For Colorize Attribute: None/Selected Attribute None is default for both
> 
> 2. Splitting in 2 different PRs - ok.

I suggest swiping controls. User selects from top to down direction. 1st - select attribute, 2nd - select mode (if needed)

ibgreen (approved):

examples/website/i3s-colorization-by-attributes/src/types.ts
```diff
@@ -0,0 +1,16 @@
+export type COLOR = [number, number, number, number];
```
belom88, line 1: CAPS is for constants

```suggestion
export type Color = [number, number, number, number];
```

examples/website/i3s-colorization-by-attributes/src/types.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type ColorsByAttribute = {
+  attributeName: string;
+  minValue: number;
+  maxValue: number;
+  minColor: COLOR;
```
belom88, line 7: There is a `Color` type in @deck.gl

belom88 (approved):

examples/website/i3s-colorization-by-attributes/src/types.ts
maxkuznetsov-actionengine: Removed type definition

examples/website/i3s-colorization-by-attributes/src/types.ts
maxkuznetsov-actionengine: added import from deck

PR #2919, opened 2024-03-15
fix(*): Fix transitive internal dependencies

### LINKED ISSUE #2918: Missing dependencies in package.json
Hello team.

I'm building an application with deck.gl and installing the latest version (9.0.0-beta.9) with npm took me to a bug here

Captura de pantalla 2024-03-15 a las 12 51 27

Seems like those two modules (`@loaders.gl/compression` and `@loaders.gl/crypto`) are not listed as explicit dependencies in the [package.json](https://github.com/visgl/loaders.gl/blob/8bafe81bc462141aab10509ac768ad7617f2f51a/modules/3d-tiles/package.json) of `@loaders.gl/3d-tiles`

Can confirm that installing those two packages in the root package.json of my project that uses deck.gl removes the bug in the screenshot above.

Thank you for your time

donmccurdy:
- fixes #2918

Missing transitive dependencies identified with:

- https://gist.github.com/donmccurdy/4f1c25d522745531ad9bebdc347e6664

I've intentionally included only the `@loaders.gl/*` dependencies here. There are others, as shown in #2918, but I'm not confident I know the effects of adding all of those.


donmccurdy:
* **#2919** <a href="https://app.graphite.dev/github/pr/visgl/loaders.gl/2919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @donmccurdy and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

modules/i3s/package.json
```diff
@@ -47,9 +47,12 @@
     "@loaders.gl/draco": "4.2.0-alpha.5",
     "@loaders.gl/images": "4.2.0-alpha.5",
     "@loaders.gl/loader-utils": "4.2.0-alpha.5",
+    "@loaders.gl/math": "4.2.0-alpha.5",
+    "@loaders.gl/polyfills": "4.2.0-alpha.5",
```
donmccurdy, line 51: Is `@loaders.gl/polyfills` meant to be in `devDependencies`, `dependencies`, or both?  Usually in a monorepo the subpackages do not have `devDependencies`, so I'm not sure whether this ...

https://github.com/search?q=repo%3Avisgl%2Floaders.gl+%40loaders.gl%2Fpolyfills+path%3Apackage.json&type=code

... is an oversight, or has a purpose I'm not aware of.

ibgreen: Polyfills should be included by client app under node.js only. No loaders module should depend on it. The module could be renamed to "node" at some point.

donmccurdy: Ok, thanks! I've removed `@loaders.gl/polyfills` from every loaders module _except_ for tile-converter, which appears to be an exception — it includes a CLI executable that depends on the polyfills.

PR #2923, opened 2024-03-21
chore: fix eslint issues

belom88:
- Enable eslint for the tile-converter module;
- Fix all lint issues;


modules/tiles/src/tileset/tileset-3d.ts
```diff
@@ -841,7 +841,7 @@ export class Tileset3D {
           break;
         default:
       }
-    } else if (this.type === TILESET_TYPE.TILES3D) {
+    } else if (this.type === 'TILES3D') {
```
mspivak-actionengine, line 844: What is the reason for this change? TILESET_TYPE enum is still being used here, e.g. line 809.
If it's decided not use enums (don't know why), it should be done across all the code, I guess.

belom88: eslint warning/error was here because`string` compared with `enum`. I guess it is considered as a potential issue

mspivak-actionengine (approved):

ibgreen (approved):

PR #2946, opened 2024-04-04
docs: update whats-new

belom88:


docs/whats-new.mdx
```diff
@@ -19,6 +20,8 @@ Target Release Date: April, 2024
 
 **@loaders.gl/parquet**
 
+- Support compression modes: 'zstd', 'snappy', 'gzip', 'brotli', 'no_compression'.
```
ibgreen, line 23: Some of these were already supported? Certainly no_compression, gzip and snappy?

docs/whats-new.mdx
```diff
@@ -19,6 +20,8 @@ Target Release Date: April, 2024
...
+- Geoparquet: support large 'Polygons' and 'MultiPolygons'. Up to 68159 vertices for 'MultiPolygon' and 20062 vertices for 'Polygon' has been tested.
```
ibgreen, line 24: It is a little odd to specify such numbers, it sounds like we don't understand what the limits in our implementation are.

ibgreen (approved):

docs/whats-new.mdx
belom88: "Geoparquet: support large 'Polygons' and 'MultiPolygons' (> 10,000 vertices)"

Is this better?

belom88: Actually, threshold was 65536 vertices (16 bit)

ibgreen: What confused me is that we reverted the Uint16 fix.

belom88: You reverted but then you've fixed it back in https://github.com/visgl/loaders.gl/pull/2940 . That was lucky change in another place )

belom88: image

belom88: So the reverted PR is not neede anymore

docs/whats-new.mdx
belom88: I don't remember right now. Daria made fixes for few types of compressions.

PR #2970, opened 2024-04-18
feat(tile-converter): Add absolute path support for i3s-server

yafedorov:


modules/tile-converter/src/i3s-server/controllers/index-controller.ts
```diff
@@ -12,9 +12,16 @@ const I3S_LAYER_PATH = process.env.I3sLayerPath || '';
  */
 export async function getFileNameByUrl(url: string): Promise<string | null> {
   const extensions = ['json', 'bin', 'jpg', 'jpeg', 'png', 'bin.dds', 'ktx2'];
-  const FULL_LAYER_PATH = path.join(process.cwd(), I3S_LAYER_PATH);
+  const absolutePath = /^[^.]+/.exec(I3S_LAYER_PATH);
```
belom88, line 15: Add a comment for regexp. It always anoying to read them.

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
```diff
@@ -12,9 +12,16 @@ const I3S_LAYER_PATH = process.env.I3sLayerPath || '';
...
+  let filePath = process.cwd();
+  let fileURLToPath = url;
+  if (absolutePath) {
+    filePath = '';
+    fileURLToPath = '';
+  }
+  const FULL_LAYER_PATH = path.join(filePath, I3S_LAYER_PATH, fileURLToPath);
   for (const ext of extensions) {
-    const fileName = `${FULL_LAYER_PATH}${url}/index.${ext}`;
+    const fileName = `${FULL_LAYER_PATH}/index.${ext}`;
```
belom88, line 24: We need to do the same for slpk

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
belom88, line 15: Does it check paths for Windows (`c:\....`)?

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
```diff
@@ -12,9 +12,16 @@ const I3S_LAYER_PATH = process.env.I3sLayerPath || '';
  */
 export async function getFileNameByUrl(url: string): Promise<string | null> {
   const extensions = ['json', 'bin', 'jpg', 'jpeg', 'png', 'bin.dds', 'ktx2'];
-  const FULL_LAYER_PATH = path.join(process.cwd(), I3S_LAYER_PATH);
+  const absolutePath = /^[^.]+/.exec(I3S_LAYER_PATH);
+  let filePath = process.cwd();
+  let fileURLToPath = url;
+  if (absolutePath) {
+    filePath = '';
+    fileURLToPath = '';
```
belom88, line 20: We discard input `url` here. It seems for me that this will work only for the root index json.

belom88 (changes requested):
Add new test cases

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
yafedorov: It should work with Windows path. Regex checks './' relative path.

ibgreen (approved):

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
belom88: Does it checkt that string starts from something. Use `I3S_LAYER_PATH.startWith('./')` instead

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
belom88: `I3S_LAYER_PATH` is the path to the http service root. What about path to the dataset? It might not be in the root. The might be multiple datasets in the root folder. 
Also there are resources - node pages, nodes. geometry, textures. All these resources should be available through http requests.

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
yafedorov: Removed regex

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
yafedorov: Added absolute path support for slpk

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
yafedorov: Added 'startWith' function to check path

modules/tile-converter/src/i3s-server/controllers/index-controller.ts
yafedorov: Corrected algorithm to have access to different resources

belom88 (approved):

PR #2999, opened 2024-05-10
feat(textures): Support loading mipmaps in KTX2 textures

donmccurdy:
Someone on the [Web Game Dev](https://www.webgamedev.com/) Discord server reported that loaders.gl doesn't return any mipmaps for KTX2 textures. Appears to be a quick fix, and the returned data looks valid. I've updated `ktx-parse` to a newer version, while I'm here.

Screenshot 2024-05-10 at 10 42 24 AM

### DIFF
--- a/modules/textures/src/lib/parsers/parse-basis.ts
+++ b/modules/textures/src/lib/parsers/parse-basis.ts
@@ -204,7 +204,6 @@ function parseKTX2File(KTX2File, data: ArrayBuffer, options): TextureLevel[][] {
 
     for (let levelIndex = 0; levelIndex < levelsCount; levelIndex++) {
       levels.push(transcodeKTX2Image(ktx2File, levelIndex, options));
-      break; // texture app can only show one level for some reason
     }
 
     return [levels];
--- a/modules/textures/package.json
+++ b/modules/textures/package.json
@@ -59,7 +59,7 @@
     "@loaders.gl/schema": "4.3.0-alpha.1",
     "@loaders.gl/worker-utils": "4.3.0-alpha.1",
     "@math.gl/types": "^4.0.1",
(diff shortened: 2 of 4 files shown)


modules/textures/src/lib/parsers/parse-basis.ts
```diff
@@ -204,7 +204,6 @@ function parseKTX2File(KTX2File, data: ArrayBuffer, options): TextureLevel[][] {
 
     for (let levelIndex = 0; levelIndex < levelsCount; levelIndex++) {
       levels.push(transcodeKTX2Image(ktx2File, levelIndex, options));
-      break; // texture app can only show one level for some reason
```
donmccurdy, line 207: @belom88 or @ibgreen do you know what "texture app" refers to here? Is that something I would need to test before merging the PR?

ibgreen: I assume this would be https://loaders.gl/examples/textures

ibgreen (approved):
Looks great, actually wasn't aware that we had this gap.

Would be good to get a stamp from @belom88 who uses this loader in the tile converter

phipno:
I tried viewing my mesh with KTX2 texture and mipmaps. But it only shows black textures.
image

ibgreen:
- Check the console for logs / errors?
- There seems to be some issues with compressed textures in luma.gl 9.1, however that is still in beta.
- Which version is I3S explorer using?
@belom88 @dsavinov-actionengine

phipno:
The console is clean, no errors.
However the textures on this page https://loaders.gl/examples/textures also seem to be black/missing.

Viewing a SLPK with KTX2 without mipmaps works though.

PR #3000, opened 2024-05-10
chore(loader-utils): Preparing to migrate from FileProviders to ReadableFiles

dariaterekhova-actionengine:


ibgreen (commented):
- Maybe outline your migration strategy in the PR comment?
- I mostly see a rename of an interface, would be good to understand how this will help the next step.

dariaterekhova-actionengine:
The main purpose of this PR is to give an instrument to use ReadableFile in already existing code (new`FileProvider` implementation) and give an opportunity to use both old FileProviders and new `ReadableFile` implementations in the same time (`FileProviderInterface`), so we can migrate one old FileProvider at a time. 

First I'm going to correct my implementation of slpk-via-rangerequests in another PR to use already existing `HttpFile` instead of new code

ibgreen (approved):

PR #3035, opened 2024-06-26
fix(mvt): Add mvt-worker.js bundle to exports path

kouak:
Ref #3034 

This MR exports the bundled worker package via `@loaders.gl/mvt/mvt-worker.js`.

I've also added an example with `vite`, and fixed a bunch of typos.


ibgreen (approved):
LGTM, and thanks for the spelling lesson :)

Pessimistress (approved):

PR #3036, opened 2024-06-26
(chore): Implement package.json#packageManager field

kouak:
Corepack appears to be the emerging tool to pin a specific package manager version.

Corepack is now distributed with node, and uses the `package.json#packageManager` field : https://nodejs.org/api/packages.html#packagemanager

This PR adds the `packageManager` field, aligned with `volta.yarn`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,5 +61,6 @@
   "volta": {
     "node": "18.19.0",
     "yarn": "4.2.2"
-  }
+  },
+  "packageManager": "yarn@4.2.2+sha512.c44e283c54e02de9d1da8687025b030078c1b9648d2895a65aab8e64225bfb7becba87e1809fc0b4b6778bbd47a1e2ab6ac647de4c5e383a53a7c17db6c3ff4b"
 }


ibgreen (commented):
Some concerns are that:
- The organization of package.json files in vis.gl repositories is aligned, so if we make a change here, we probably should be making it across all repos (deck.gl, luma.gl, math.gl, etc etc)
- We already use volta to control package manager, so we are adding a separate mechanism for node only. In what cases is this beneficial?
- When someone upgrades the yarn version, how do they find the SHAs of the new version?
- And finally what does this actually control? How is node involved in package management? What node command delegates to the package manager specified here.

kouak:
corepack replaces the package manager part of volta

If all team members use volta and are fine with it, then this is redundant. This is beneficial to external contributors who might not have volta installed on their system but are more likely to have corepack enabled on their system IMO (since corepack is distributed alongside node).

When enabling corepack on a system (via `corepack enable`), corepack will create `pnpm` and `yarn` binaries and intercept all calls.

When `yarn` or `pnpm` is called in a project, corepack will automatically download and cache the proper package manager version :

image

To update to a specific `yarn` version, maintainers should run `corepack use yarn@4.2.2` in the repo.

To update to the latest `yarn` version : `corepack up`.

On corepack enabled systems, corepack will try to set the `package.json#packageManager` field if its not present.

Corepack works fine without the sha  :
```json
{
  "packageManager": "yarn@4.2.2"
}
```

There's an [upcoming vote](https://github.com/nodejs/TSC/pull/1527) in Node Technical Steering Committee about whether future node versions should enable corepack by default (and thus, install `yarn` and `pnpm` binaries, managed by corepack).

We just migrated our team at work from volta to corepack + [fnm](https://github.com/Schniz/fnm) and we've found the developer experience to be similar or better. We're building a dockerized app though, and because corepack is bundled with node, corepack is also present in official node docker images. In our case, this ensures that our artefacts also use the same package manager version.

This might not be relevant to the vis.gl repos though, so feel free to close this PR in that case.

PR #3051, opened 2024-07-30
fix(terrain): Add terrain-worker.js bundle to exports path

attilam:
Ref #3035 

I have added an export to the terrain-worker, since I needed that.

Thanks!


ibgreen (approved):

attilam:
Sorry, missing comma! 🥶

PR #3065, opened 2024-08-22
chore(potree): test PotreeSource on a supported dataset

belom88:


ibgreen (commented):
Looks like quite a bit of data to add to the repo. Can we include just a subset or add this to e.g. deck.gl-data?

modules/potree/test/potree-source.spec.ts
```diff
@@ -1,11 +1,13 @@
 import test from 'tape-promise/tape';
 import {PotreeSource} from '@loaders.gl/potree';
 
-const POTREE_HIERARCHY_CHUNK_URL = '@loaders.gl/potree/test/data/lion_takanawa';
+const POTREE_BIN_URL = '@loaders.gl/potree/test/data/lion_takanawa';
+const POTREE_LAZ_URL =
```
ibgreen, line 5: Maybe use version numbers here? POTREE_V17_URL, POTREE_V18_URL etc?

ibgreen (approved):
- Maybe add `lion_takanawa` to deck.gl-data as 1.7 example (we can still use the local copy for speed).
- Ideally we should have a full complement of versions 1.6 2.0 in deck.gl-data.

modules/potree/test/potree-source.spec.ts
belom88: The criteria for these datasets here is content type (BIN/LAZ). 1.7 and 1.8 are almost the same. The significant difference that we don't support bin content a the moment.

PR #3086, opened 2024-09-16
fix(3d-tiles): Rewrite of parseImplicitTiles for readability

olsen232:
Fixes https://github.com/visgl/loaders.gl/issues/3085.
It's possible to fix that issue with smaller changes, but I believe I found and fixed some more bugs by rewriting it for clarity.
Adds comments.

### DIFF
--- a/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-implicit-tiles.ts
+++ b/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-implicit-tiles.ts
@@ -79,47 +79,46 @@ function getChildS2VolumeBox(
  * Recursively parse implicit tiles tree
  * Spec - https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_implicit_tiling
  * TODO Check out do we able to use Tile3D class as return type here.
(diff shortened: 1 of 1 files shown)


belom88 (commented):
Thank you for this fix @olsen232 .
Regarding https://github.com/visgl/loaders.gl/issues/3085#issuecomment-2352034069 : you can build code on linux or mac os. Use wsl on Windows. You have a ts error right now.
We will check this fix on the tile-converter.

ibgreen (approved):
@olsen232 Appreciate the contribution. the changes are fine from a high-level perspective, since they are localized inside existing function.  

I will rely on @belom88 for the detailed review of the logic.

Do we have test cases for this? I think we need at least something to make sure future maintainers can catch regressions.

Not sure why our contributing page is not part of the official docs: https://github.com/visgl/loaders.gl/blob/eb084379121e8bbf4d4367c3b380c15f2fd646bf/CONTRIBUTING.md#L4

belom88 (approved):

belom88:
Tested on a dataset with quadtree implicit tiles (CDB_Yemen_elevation)

PR #3092, opened 2024-09-19
added ktx2 to the list of file extensions

mspivak-actionengine:


ibgreen (commented):
Change looks fine, but can you separate out the yarn.lock update?

dsavinov-actionengine:
> Change looks fine, but can you separate out the yarn.lock update?

@ibgreen it seems without the yarn.lock update the tests on this PR are failing. We can merge the yarn.lock PR 1st and then this one @mspivak-actionengine

PR #3163, opened 2024-11-19
fix(): Hofix to remove not existing version

dariaterekhova-actionengine:
I found out that out i3s explorer pulls some code from beta versions and it currently breaks it. This change is a hot fix for the i3s explorer to continue working at the moment.


ibgreen (approved):
Hmm, I thought would be auto-updated during publish....

dariaterekhova-actionengine:
No, lerna doesn't update it. That's why it was some default stable version previously. And I'll put it back in the future probably

PR #3194, opened 2025-03-09
fix(3d-tiles): fix loading tilesets with `availableLevels` = `1` in the implicit tiling scheme

### LINKED ISSUE #3193: [3d-tiles] `parseImplicitTiles` fails to load subtree if `maximumLevel` is `0`
The condition in the line below fails to account for the case when `maximumLevel` is `0`:

https://github.com/visgl/loaders.gl/blob/d74df445180408fd772567ca6a6bbb8d42aa50be/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-implicit-tiles.ts#L139

I'm admittedly no 3d tiles expert but I guess this should be `!Number.isFinite(maximumLevel)` in order to cater for tilesets where the implicit tiling scheme defines a single level (copy-pasting the following from a Cesium Tileset we're using):

```json
    "implicitTiling": {
      "subdivisionScheme": "OCTREE",
      "subtreeLevels": 5,
      "availableLevels": 1,
      "subtrees": { "uri": "subtrees/{level}/{x}/{y}/{z}.subtree" }
    },
```

`maximumLevel` is being set to `0` here in `normalizeImplicitTileHeaders()`:

https://github.com/visgl/loaders.gl/blob/d74df445180408fd772567ca6a6bbb8d42aa50be/modules/3d-tiles/src/lib/parsers/parse-3d-tile-header.ts#L226

brncsk:
Fixes #3193

### DIFF
--- a/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-implicit-tiles.ts
+++ b/modules/3d-tiles/src/lib/parsers/helpers/parse-3d-implicit-tiles.ts
@@ -136,7 +136,7 @@ export async function parseImplicitTiles(params: {
   } = implicitOptions;
   const tile = {children: [], lodMetricValue: 0, contentUrl: ''};
 
-  if (!maximumLevel) {
+  if (maximumLevel === undefined) {
     log.once(
       `Missing 'maximumLevel' or 'availableLevels' property. The subtree ${contentUrlTemplate} won't be loaded...`
     );


ibgreen:
Thanks for the fix. This is now superseded by the stronger validation on `master`:

```ts
if (typeof maximumLevel !== 'number' || !Number.isFinite(maximumLevel)) {
```

That accepts `maximumLevel === 0` (the `availableLevels === 1` case) while also rejecting missing and non-finite values. Closing as no longer needed.

PR #3272, opened 2025-12-30
Remove caution about LAS v1.4 file support

athomann:
Removed caution note about LAS v1.4 file support. 1.4 support added with https://github.com/visgl/loaders.gl/pull/3165


sguergachi:
I still see the notice https://loaders.gl/docs/modules/las 
image

Also on https://loaders.gl/docs/modules/las/formats/las it still shows 1.4 not updated 
image

PR #3311, opened 2026-03-03
Fix(tiles): Remove rendering flicker by holding refined tile

felixpalmer:
Fixes https://github.com/visgl/deck.gl/issues/7914 (requires also https://github.com/visgl/deck.gl/pull/10058)

### Background

As the 3D tiles traversal algorithm loads/unloads tiles the data presented to the rendered (e.g. deck) changes very often. Flashes often appear as tiles are removed from the render list, before those meant to replace them are initialized and drawn.

This PR introduces a mechanism for the `Tileset3D` to temporarily "hold onto" tiles which are about to disappear while the renderer has tiles which are "selected" (i.e. marked for drawing) but haven't yet been drawn.

The fix works in conjunction with changes in deck.gl and is bi-directionally back-compatible, falling back gracefully to the old behavior if the renderer doesn't support it.

#### Before

https://github.com/user-attachments/assets/877be080-fd25-447a-a136-4929bcc52a38

#### After

https://github.com/user-attachments/assets/bb9ff0a4-881f-494e-9d9f-952ba3b466e5

### Changes

- Add `tileDrawn` flag to `Tile3D` class
- Track if any tiles are marked as undrawn in Tileset3D and prevent any tiles being removed during this (brief) window
- Add Tileset3D.onUpdate() hook to allow signalling that a redraw is needed in the renderer

CC @akre54 @charlieforward9

### DIFF
--- a/modules/tiles/test/tileset/tile-3d.spec.ts
+++ b/modules/tiles/test/tileset/tile-3d.spec.ts
@@ -301,6 +301,20 @@ test('Tile3D#viewerRequestVolume is camera outside the OBB viewer request volume
   t.end();
 });
 
+test('Tile3D#tileDrawn defaults to true', (t) => {
+  // @ts-ignore
+  const tile = new Tile3D(MOCK_TILESET, TILE_HEADER_WITH_BOUNDING_SPHERE);
+  t.equals(tile.tileDrawn, true, 'tileDrawn defaults to true for backwards compatibility');
+
+  tile.tileDrawn = false;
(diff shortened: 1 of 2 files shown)


akre54:
Thanks @felixpalmer !

It also might be nice to support fading between tiles as they load, but assuming that can be done in a follow-up (I see a few hooks here that look possibly relevant)

charlieforward9 (approved):
Clean! 

Could 2D tiles need this as well? 

I've certainly seen some resolution issues before and I am curious if logic like this would be part of some common `Tile`-class?

PR #3335, opened 2026-03-29
[example] select url from tileset param

### LINKED ISSUE #3333: [Bug] 3d-tiles example url parameter option does not work
### Loader

3d-tiles example

### Description

The 3d-tiles example has a great url parameter option to load custom tilesets. It currently does not work. For example

https://loaders.gl/examples/3d-tiles/?tileset=https://storage.googleapis.com/open-cogs/planet-stac/angkor-wat/3d-geofox.ai/3DTiles/tileset.json

zooms to the correct place but does not render the tileset:

Image

### Expected Behavior

https://loaders.gl/examples/3d-tiles/?tileset=https://storage.googleapis.com/open-cogs/planet-stac/angkor-wat/3d-geofox.ai/3DTiles/tileset.json

should show

Image

### Steps to Reproduce

see above

### Environment

- Framework version: current
- Browser: current Chrome
- Node:
- OS: Win11

### Logs

None

andreasplesch:
fixes #3333

The change to `trailingSlash: undefined` was necessary for serving from the developmental docusaurus server in order to keep the url unchanged by the server. It may not be necessary for serving from the published loaders.gl website.

### DIFF
--- a/examples/website/3d-tiles/app.tsx
+++ b/examples/website/3d-tiles/app.tsx
@@ -99,8 +99,6 @@ export default class App extends PureComponent {
     this._tilesetStatsWidget = new StatsWidget(null, {container});
 
     this._loadExampleIndex();
-
-    
   }
 
   // load the index file that lists example tilesets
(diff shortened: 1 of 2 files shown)


ibgreen:
Thanks for this. The primary fix is already on `master`: `_selectTilesetFromQueryParams()` is now called from `_loadExampleIndex()` after `examplesByCategory` has loaded, so URL-selected tilesets no longer race initialization.

The remaining `trailingSlash` change is independent and was noted as potentially unnecessary for the published site. If that setting is still needed, please open a narrowly scoped follow-up with a reproducible case. Closing this PR as superseded.

PR #3432, opened 2026-05-18
Fix tile LOD transitions: improve REPLACE refinement handling

olsen232:
### Summary

Replace aggressive tile holding mechanism with precise REPLACE-only transitions:
- Remove global _heldTiles set that was holding all previously selected tiles
- Only hold parent tiles during REPLACE refinement when their selected children haven't drawn yet
- Prevents low LOD tiles from inappropriately blocking high LOD models
- Maintains smooth transitions without single-frame flashes

### Background

I found that from loaders.gl 9.3 onwards, LOD refinement wasn't working properly, at least on certain datasets.
The culprit seems to be https://github.com/visgl/loaders.gl/commit/bc88c2ebaaecdca9ab264ca203d083b78998bd1a
That commit does indeed get rid of the white flashes that happen when lower-LOD tiles are dropped before higher LOD tiles are ready, but it seems to be too aggressive in holding onto lower-LOD tiles. For the dataset I've been testing with, the lower LOD tiles never go away and are shown alongside the higher LODs, sometime occluding them and sometimes not, giving a mottled, mixed-LOD view.

Making the logic stateless - getting rid of the longer-lasting _heldTiles variable and instead just expanding the list of selectedTiles somewhat right after the list of selectedTiles is generated - seems to fix the issue.

Unfortunately, I don't own the datasets I've been testing with so I can't attach them to the PR - hopefully you can find another dataset with the same issue, let me know how this goes.

### Before this fix:

before-fix

### After this fix:

after-fix

### DIFF
--- a/modules/tiles/src/tileset-3d/common/tileset-3d.ts
+++ b/modules/tiles/src/tileset-3d/common/tileset-3d.ts
@@ -14,7 +14,7 @@ import {FrameState, getFrameState, limitSelectedTiles} from '../helpers/frame-st
 
 import type {GeospatialViewport, Viewport} from '../../types';
 import {Tile3D} from './tile-3d';
-import {TILESET_TYPE} from '../../constants';
+import {TILESET_TYPE, TILE_REFINEMENT} from '../../constants';
 
 import {TilesetTraverser} from './tileset-traverser';
 import {
@@ -200,7 +200,6 @@ export class Tileset3D {
   _cache = new TilesetCache();
   _requestScheduler: RequestScheduler;
 
-  private _heldTiles: Set<string> = new Set();
   private updatePromise: Promise<number> | null = null;
   tilesetInitializationPromise: Promise<void>;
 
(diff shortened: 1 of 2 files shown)


charlieforward9:
This is a positive fix I am looking for in my use case. Would you be able to crop the vegetation from the tileset and pass just that as a reproduction case for this to land sooner? Thank you

olsen232:
[WaikatoFarmBoxes.zip](https://github.com/user-attachments/files/28979964/WaikatoFarmBoxes.zip)

I didn't attach the original WaikatoFarm dataset that I screenshotted above but I did attach a 3dtiles that has all of the same tileset.json - in place of actual lidar scans of the Waikato, it just has wireframe boxes. Red wireframes are low LOD, orange then yellow wireframes are higher LOD.
The original 3dtiles was constructed by software from https://www.skylinesoft.com/

I guess verifying that this fix does something good would involve:

- checking that it doesn't cause regressions with other 3dtiles you do have access to: they render with appropriate LOD, no white flashes
- before this fix is applied, red wireframe boxes have a tendency to hang around even when yellower are visible
- after this fix is applied, red wireframe boxes have less of a tendency to hang around even when yellower are visible

I'll try and do the same - figuring out how to test this end to end always takes me a while, so I've attached this already while I do this again.

olsen232:
I have been struggling to make WaikatoFarmBoxes display the same issue as the original - even though they share the same structure, I can visibly reproduce the issue with the actual farm dataset, or not reproduce the issue if I got to a different version of loaders.gl - but the colourful boxes appear roughly the same in either case.

I even tried to increase the size in bytes of the colourful boxes to make them the same size as the originals so that it wouldn't make a difference, but of course it's just empty padding so maybe it made no difference.

Anyway even though I can still see this issue, I'm having a lot of difficulty making a repro case that I have ownership to share.

olsen232:
🫤 OK I think I've been doing something else at the same time that's been making this issue worse for me. I still think the fix from this PR is somewhat good but it's usefulness may be a lot more limited than I realised.

The WaikatoFarm dataset that I've been trying to attach here without actually sharing has an extension in it called SKYLINE_floor - I don't know what this extension is supposed to do, but it has some extra tile data that, when rendered naively, makes the dataset look worse. I've been using a patched loader that discards any GLTF that uses the SKYLINE_floor extension for this reason, but I hadn't realised that it's probably making this bug worse.

By doing this, I think I'm effectively simulating a case where some tiles are effectively corrupt and never load. In such a case, the current hold-tile-while-waiting-for-next-LOD behaviour is indeed more aggressive than needed, and I was able to replace it with a less aggressive variant that still avoids the white flashes, but doesn't hold onto low LOD tiles forever.

In the WaikatoFarm dataset: notably the extensions is are in a separate tree called 1/ and the actual renderable data tree is 2/. The failure of certain tiles to load in tree 1/ shouldn't affect how one LOD is replaced with another, or not, in tree 2. So, I think I have found a bug - or at least something that can be improved - it's just hard to reproduce without preventing certain tiles from loading.


### tl;dr:
I believe this fix makes the hold-tile-while-waiting-for-next-LOD logic simpler and more logical: tiles are held only while their children are not quite ready to draw, there's no global state that's held long term.
However, the difference between the current logic and my proposed logic will not be noticeable by most people at most times - I think I only came across it because I was preventing certain tiles from loading.


### Next steps
It's up to you how you want to proceed here.
- you could just merge this since it makes the code slightly tidier
- "This is a positive fix I am looking for in my use case." I'm not sure exactly what this means, but if this fix does something you need, then I guess you'll want to try and proceed with it. But if so, you'll have to check this fix against your own use-case, since my use-case is a bit mixed up at this point.

If you decide to close this instead, I'll have to find a different patch for the SKYLINE_floor extension that doesn't interact poorly with this not-quite-ideal LOD logic. That's okay too I guess, but leaving it up to you since I believe this PR is still a mild improvement to the codebase.

PR #3463, opened 2026-07-11
Upgrade Parquet Thrift dependency to 0.23

evcatalyst:
Updates the Parquet module from Thrift 0.19 to 0.23 so npm consumers receive the safe dependency without relying on a Yarn-only resolution. The Parquet Node suite passes 82 tests with 4 skipped.


ibgreen-openai:
Closed in separate PR

PR #3476, opened 2026-07-29
fix(tiles): Add groundHeightDatum option to fix culling of high-altitude region tilesets

### LINKED ISSUE #3475: 3D Tiles: all content tiles culled for high-altitude regions — getFrameState treats height-above-map-plane as height-above-ellipsoid
### Summary

`Tileset3D` (and therefore deck.gl's `Tile3DLayer`) culls **all content tiles** of a 3D Tiles dataset whose geometry sits at real altitude (e.g. Switzerland, ~400–1500 m a.s.l.) whenever the map viewport is zoomed in to street level. The tileset-JSON hierarchy loads, but no `.b3dm` content is ever requested — the layer silently renders nothing. Zooming far out makes the same dataset load fine.

Root cause: `getFrameState()` builds the culling camera and all six frustum planes by feeding `viewport.unprojectPosition(point)` — whose z is **metres above deck's z=0 map plane** — directly into `Ellipsoid.WGS84.cartographicToCartesian(...)`, which interprets that z as **metres above the WGS84 ellipsoid**. No terrain/ground datum exists anywhere in `modules/tiles/src/tileset-3d/helpers/frame-state.ts` (`worldToCartesian`, camera position, `commonSpacePlanesToWGS84` all share the conversion). The culling frustum is therefore constructed as if the scene sat at sea level. Content bounding `region` volumes are absolute per the 3D Tiles spec (they ignore tile/tileset transforms), so for high-altitude content the frustum is built hundreds of metres too low: zoomed out it is sloppy enough to still intersect the volumes; at street zoom the tight leaf volumes (e.g. `hmin≈400, hmax≈480`) fall outside and are culled before any content request. Demos over near-sea-level cities (SF/NYC) never hit this, which is presumably why it has gone unnoticed.

Because region volumes ignore transforms, **no `modelMatrix` workaround exists** — we verified that translating the tileset only moves the rendered mesh, never the culling volumes.

### Environment

- `@loaders.gl/tiles` / `@loaders.gl/3d-tiles` 4.4.3, `@deck.gl/core` 9.3.7, node v22
- Dataset: swisstopo swissBUILDINGS3D (public open data, classic 1.0 tileset with nested external tilesets and `region` bounding volumes at real Swiss altitudes)
- Also reproduced in-browser via deck.gl `Tile3DLayer` in a MapLibre `MapboxOverlay` (interleaved) — identical symptom: tileset JSONs fetch, zero b3dm requests, regardless of camera interaction.

### Minimal repro (node, no WebGL, no browser)

```bash
npm init -y && npm pkg set type=module
npm i @deck.gl/core @loaders.gl/tiles @loaders.gl/3d-tiles @loaders.gl/core @math.gl/core
node repro.mjs
```

```js
// Repro: Tileset3D culls all content tiles for high-altitude regions.
//
// getFrameState() (modules/tiles/src/tileset-3d/helpers/frame-state.ts) converts the
// deck.gl viewport camera + frustum planes to ECEF via viewport.unprojectPosition(),
// whose z is metres above deck's z=0 MAP PLANE — but cartographicToCartesian()
// interprets that z as metres above the WGS84 ELLIPSOID. There is no terrain datum.
// For content whose (spec-absolute) region bounding volumes sit at real altitude
// (Swiss buildings: ~400-480 m a.s.l.), the culling frustum is built ~400 m too low:
// zoomed out it still intersects; at street zoom every leaf volume is culled and no
// content is ever requested.
//
// Run: node repro.mjs            (runs both viewports, prints a summary)
// Public dataset: swisstopo swissBUILDINGS3D (open data).
// NOTE: b3dm PARSE errors about a missing draco wasm are irrelevant noise in node —
// the bug is about whether content tiles are ever REQUESTED at all.

import {WebMercatorViewport} from '@deck.gl/core';
import {Tileset3D} from '@loaders.gl/tiles';
import {Tiles3DLoader} from '@loaders.gl/3d-tiles';
import {load} from '@loaders.gl/core';

const TILESET = 'https://3d.geo.admin.ch/ch.swisstopo.swissbuildings3d.3d/v1/tileset.json';
const ZURICH = {longitude: 8.5391, latitude: 47.3686}; // Bahnhofstrasse, Zürich (~408 m a.s.l.)

async function run(label, {width, height, zoom}) {
  const fetched = [];
  const origFetch = globalThis.fetch;
  globalThis.fetch = async (u, o) => {
    fetched.push(String(u));
    return origFetch(u, o);
  };

  const viewport = new WebMercatorViewport({
    id: 'repro', width, height, ...ZURICH, zoom, pitch: 0, bearing: 0,
  });

  const tilesetJson = await load(TILESET, Tiles3DLoader);
  const tileset = new Tileset3D(tilesetJson, {
    onTileError: () => {}, // draco wasm noise in node; requests are what matter
  });

  // The tileset is a deep chain of external tilesets; each selectTiles round
  // descends one hop (mirrors deck.gl's viewportChanged/onTileLoad update cycle).
  for (let round = 1; round <= 8; round++) {
    await tileset.selectTiles([viewport]);
    await new Promise((r) => setTimeout(r, 2000));
  }

  const b3dm = fetched.filter((u) => u.includes('.b3dm'));

  // Visibility probe: walk the per-viewport tree, find a leaf content tile near
  // Zürich, and report visibility up its ancestor chain.
  const all = [];
  const walk = (t) => { all.push(t); (t.children || []).forEach(walk); };
  for (const r of Object.values(tileset.roots || {})) walk(r);
  const leaf = all.find((t) => /\.b3dm/.test(String(t.id || '')) &&
    inRegion(t.header?.boundingVolume?.region, ZURICH));
  let probe = 'no Zürich leaf tile in tree (traversal never spliced that deep)';
  if (leaf) {
    const chainTail = [];
    for (let n = leaf, i = 0; n && i < 4; n = n.parent, i++) chainTail.push(n);
    probe = chainTail.map((t) => {
      const reg = t.header?.boundingVolume?.region;
      const h = reg ? `hmin=${Math.round(reg[4])} hmax=${Math.round(reg[5])}` : 'noregion';
      return `vis=${t._visible} dist=${Math.round(t._distanceToCamera ?? -1)} ${h}`;
    }).join('  <-  ');
  }

  globalThis.fetch = origFetch;
  console.log(`\n=== ${label} (${width}x${height} zoom ${zoom})`);
  console.log(`tileset JSONs fetched: ${fetched.filter((u) => u.includes('.json')).length}`);
  console.log(`.b3dm content tiles REQUESTED: ${b3dm.length}${b3dm.length ? '  e.g. ' + b3dm[0].split('/').slice(-3).join('/') : '  <-- BUG: traversal stalled'}`);
  console.log(`leaf->root visibility (leaf first): ${probe}`);
  return b3dm.length;
}

function inRegion(region, {longitude, latitude}) {
  if (!region) return false;
  const [w, s, e, n] = region;
  const lon = (longitude * Math.PI) / 180;
  const lat = (latitude * Math.PI) / 180;
  return w <= lon && lon <= e && s <= lat && lat <= n;
}

const wide = await run('zoomed out — WORKS', {width: 1200, height: 700, zoom: 16.5});
const tight = await run('street zoom — STALLS', {width: 898, height: 320, zoom: 17});
console.log(`\nSUMMARY: zoomed-out b3dm requests = ${wide}; street-zoom b3dm requests = ${tight}`);
console.log(tight === 0 && wide > 0
  ? 'REPRODUCED: identical tileset + identical code path; only the viewport differs.'
  : 'not reproduced in this environment');
```

### Observed output

```
=== zoomed out — WORKS (1200x700 zoom 16.5)
tileset JSONs fetched: 5
.b3dm content tiles REQUESTED: 6  e.g. 11/1165/1601.b3dm
leaf->root visibility (leaf first): vis=true dist=162 hmin=405 hmax=438  <-  vis=true dist=159 hmin=403 hmax=441  <-  vis=true dist=115 hmin=398 hmax=485  <-  vis=true dist=107 hmin=395 hmax=493

=== street zoom — STALLS (898x320 zoom 17)
tileset JSONs fetched: 5
.b3dm content tiles REQUESTED: 0  <-- BUG: traversal stalled
leaf->root visibility (leaf first): vis=undefined dist=0 hmin=405 hmax=438  <-  vis=undefined dist=0 hmin=403 hmax=441  <-  vis=undefined dist=0 hmin=398 hmax=485  <-  vis=undefined dist=0 hmin=395 hmax=493

SUMMARY: zoomed-out b3dm requests = 6; street-zoom b3dm requests = 0
REPRODUCED: identical tileset + identical code path; only the viewport differs.
```

(`vis=undefined` means the traverser never evaluated those tiles because an ancestor's children were culled during `updateAndPushChildren`; calling `tile.updateVisibility(frameState)` manually on the leaf in the stalled viewport returns `_visible=false` with camera-to-volume distance ≈ 0–1 m — the frame places the camera *inside* the buildings' altitude band.)

### Suggested fix

Add a height-datum term to the viewport→ECEF conversions in `frame-state.ts` — either an explicit tileset/frame option (e.g. `groundHeightOffset`), or auto-derived from the root tile's `region` `minimumHeight` — and apply it to the z of the camera cartographic and every `worldToCartesian` call (ENU origin + the six culling-plane points). That aligns the culling frame with where the content actually is, matching what globe engines (which keep camera and tiles in one ECEF frame) get for free.

Happy to provide more traces — we hit this in production over Zürich and diagnosed it down to these lines.

Mkuesta:
Fixes #3475

## Goals

`Tileset3D` (and therefore deck.gl's `Tile3DLayer`) culls **all content tiles** of a 3D Tiles dataset whose geometry sits at real altitude (e.g. Switzerland, ~400–1500 m a.s.l.) whenever the viewport is zoomed to street level — the tileset JSON hierarchy loads, but no `.b3dm` is ever requested.

Root cause (diagnosed in #3475): `getFrameState()` builds the culling camera and all six frustum planes by feeding `viewport.unprojectPosition(point)` — whose z is metres above deck's z=0 map plane — directly into `Ellipsoid.WGS84.cartographicToCartesian(...)`, which interprets that z as metres above the WGS84 ellipsoid. No ground datum exists, so the culling frame is always built at sea level. `region` bounding volumes are spec-absolute (they ignore transforms), so for high-altitude content the tight leaf volumes fall outside the street-zoom frustum and are culled before any content request. No `modelMatrix` workaround exists.

This PR aligns the culling frame with the content by introducing a **ground-height datum**: the elevation of the viewport's ground plane above the ellipsoid, applied as a uniform vertical re-anchor of the whole culling frame — auto-derived from the tileset's `region` volumes by default, overridable via a new public option.

## Changes

- **`modules/tiles/src/tileset-3d/helpers/frame-state.ts`**
  - `getFrameState(viewport, frameNumber, options?)` — new optional third parameter (`GetFrameStateOptions`) with `groundHeightDatum` (metres). The datum is added to the z of the camera cartographic and every `worldToCartesian` conversion (ENU origin + all six culling-plane anchor points). Default `0` → output identical to before (backward compatible for the existing 2-arg public call signature).
  - New exported `resolveGroundHeightDatum(option, rootTile, viewportCenter)` — resolves the tileset option: an explicit number wins; `'auto'` derives the minimum height of the **deepest loaded `region` containing the viewport center**, seeded by the root region's `minimumHeight`. The tile tree deepens as external tilesets load, so successive traversal frames converge on the local ground height (root-region-only derivation is not sufficient for country-scale tilesets: the swisstopo root spans 149–4564 m while Zürich's local ground is ~408 m). Tilesets without `region` volumes (box/sphere/I3S) resolve to `0` → exact no-op.
- **`modules/tiles/src/tileset-3d/common/tileset-3d.ts`** — new `groundHeightDatum?: number | 'auto'` option (default `'auto'`), resolved per viewport per frame in `doUpdate()`, so `tileset.setProps({groundHeightDatum})` works at runtime — that is also the override path for deck.gl `Tile3DLayer` users: `onTilesetLoad: (tileset) => tileset.setProps({groundHeightDatum: 408})`.
- **`modules/tiles/src/index.ts`** — export `resolveGroundHeightDatum` + `GetFrameStateOptions`.
- **`docs/modules/tiles/api-reference/tileset-3d.md`** — option documented.
- **Tests**
  - `modules/tiles/test/tileset/helpers/ground-height-datum.spec.ts` (new, node + browser): resolver unit tests — explicit override (incl. explicit `0` to disable), root-region derivation, deepest-containing-region refinement, non-region fallbacks, cyclic-graph termination.
  - `modules/tiles/test/tileset/helpers/get-frame-state.spec.ts`: datum shifts the camera cartographic height by exactly the datum with lon/lat unchanged and all six planes lifted; regression test reproducing #3475 (Zürich street-zoom viewport + a 405–438 m region OBB: culled at sea level, kept with the local-ground datum).
  - `modules/tiles/test/tileset/tileset-3d-traversal.spec.ts`: option wiring — an extreme datum culls the whole (sea-level) fixture; default options leave it visible.

## Verification

The issue's standalone node repro (adapted only for the 5.0 `Tiles3DSource` constructor), swisstopo swissBUILDINGS3D over Zürich, **no options passed** (pure default `'auto'`):

Before (master `f7a4b32`):

```
SUMMARY: zoomed-out b3dm requests = 6; street-zoom b3dm requests = 0
REPRODUCED: identical tileset + identical code path; only the viewport differs.
```

After:

```
=== street zoom (898x320 zoom 17)
.b3dm content tiles REQUESTED: 6  e.g. 11/1165/1601.b3dm
leaf->root visibility (leaf first): vis=true dist=161 hmin=405 hmax=438  <-  vis=true dist=158 ...
SUMMARY: zoomed-out b3dm requests = 20; street-zoom b3dm requests = 6
not reproduced in this environment
```

`yarn test-node` and `yarn test-headless` fully green; `yarn build` and `yarn lint fix` clean.

## Behavior notes

- Sea-level `region` tilesets (minimumHeight ≈ 0), box/sphere tilesets and I3S resolve to datum 0 — default-path output is unchanged (the untouched existing suites are the regression net).
- High-altitude `region` tilesets see two intended changes: content stops being wrongly culled, and camera-to-tile distances now match the content frame, so SSE refines slightly more (the zoomed-out Zürich viewport goes from 6 to 20 b3dm requests because distances are no longer inflated by the missing ~400 m).
- `'auto'` is best-effort and converges as region headers load; for degenerate cases (extreme relief inside one leaf region, unusual viewports) the explicit option is the escape hatch.
- I3S's `getProjectedRadius` performs its own datum-less conversion; `'auto'` resolves to 0 for I3S so I3S defaults are consistent (pre-existing gap, out of scope here).
- Happy to backport to `4.4-release` if wanted — we hit this in production on 4.4.3.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/modules/tiles/api-reference/tileset-3d.md
+++ b/docs/modules/tiles/api-reference/tileset-3d.md
@@ -78,6 +78,7 @@ Parameters:
   - `options.loadOptions` - _loaders.gl_ options used when loading tiles from the tiling server. Includes `fetch` options such as authentication `headers`, worker options such as `maxConcurrency`, and options to other loaders such as `3d-tiles`, `gltf`, and `draco`.
   - `options.contentLoader` = `null` (`Promise`) - An optional external async content loader for the tile. Once the promise resolves, a tile is regarded as _READY_ to be displayed on the viewport.
   - `options.loadTiles`=`true` (`Boolean`) - Whether the tileset traverse and update tiles. Set this options to `false` during the run time to freeze the scene.
+  - `options.groundHeightDatum`=`'auto'` (`Number | 'auto'`) - Elevation in meters of the viewport's ground plane (z = 0) above the WGS84 ellipsoid, used to align the culling frustum with content whose bounding volumes carry absolute heights (3D Tiles `region` volumes at high altitude, e.g. alpine cities). `'auto'` derives it from the minimum height of the deepest loaded `region` containing the viewport center. Set an explicit number to override (e.g. the terrain height at the camera target), or `0` to disable. With deck.gl's `Tile3DLayer`, override via `onTilesetLoad: (tileset) => tileset.setProps({groundHeightDatum: 408})`.
 
 Callbacks:
 
--- a/modules/tiles/src/index.ts
+++ b/modules/tiles/src/index.ts
@@ -83,7 +83,8 @@ export {PointCloudTile} from './point-cloud/point-cloud-tile';
 export {createBoundingVolume} from './tileset-3d/helpers/bounding-volume';
 export {calculateTransformProps} from './tileset-3d/helpers/transform-utils';
 
-export {getFrameState} from './tileset-3d/helpers/frame-state';
+export {getFrameState, resolveGroundHeightDatum} from './tileset-3d/helpers/frame-state';
+export type {GetFrameStateOptions} from './tileset-3d/helpers/frame-state';
 export {getLodStatus} from './tileset-3d/helpers/i3s-lod';
 
 export {
--- a/modules/tiles/src/tileset-3d/common/tileset-3d.ts
+++ b/modules/tiles/src/tileset-3d/common/tileset-3d.ts
@@ -10,7 +10,12 @@ import {Stats} from '@probe.gl/stats';
 import {RequestScheduler, LoaderWithParser, LoaderOptions} from '@loaders.gl/loader-utils';
 import {TilesetCache} from './tileset-cache';
 import {calculateTransformProps} from '../helpers/transform-utils';
-import {FrameState, getFrameState, limitSelectedTiles} from '../helpers/frame-state';
+import {
+  FrameState,
+  getFrameState,
+  limitSelectedTiles,
+  resolveGroundHeightDatum
+} from '../helpers/frame-state';
 
 import type {GeospatialViewport, Viewport} from '../../types';
 import {Tile3D} from './tile-3d';
@@ -50,6 +55,15 @@ export type Tileset3DProps = {
   viewportTraversersMap?: any;
   updateTransforms?: boolean;
   viewDistanceScale?: number;
+  /**
+   * Elevation in meters of the viewport's ground plane (z = 0) above the WGS84 ellipsoid,
+   * used to align the culling frustum with content whose bounding volumes carry absolute
+   * heights (3D Tiles `region` volumes). `'auto'` (default) derives it from the minimum
+   * height of the deepest loaded `region` containing the viewport center (seeded by the
+   * root region); tilesets without region volumes resolve to 0. Set an explicit number
+   * to override, or 0 to disable.
+   */
+  groundHeightDatum?: number | 'auto';
 
   onTileLoad?: (tile: Tile3D) => any;
   onTileUnload?: (tile: Tile3D) => any;
@@ -83,6 +97,7 @@ type Props = {
   maximumScreenSpaceError: number;
   memoryAdjustedScreenSpaceError: boolean;
   viewportTraversersMap: Record<string, any> | null;
+  groundHeightDatum: number | 'auto';
   attributions: string[];
   loadTiles: boolean;
   loadOptions: LoaderOptions;
@@ -117,6 +132,7 @@ const DEFAULT_PROPS: Props = {
   loadTiles: true,
   updateTransforms: true,
   viewportTraversersMap: null,
+  groundHeightDatum: 'auto',
   loadOptions: {fetch: {}},
   attributions: [],
   basePath: '',
@@ -354,7 +370,13 @@ export class Tileset3D {
       if (!viewportsToTraverse.includes(id)) {
         continue;
       }
-      const frameState = getFrameState(viewport as GeospatialViewport, this._frameNumber);
+      const frameState = getFrameState(viewport as GeospatialViewport, this._frameNumber, {
+        groundHeightDatum: resolveGroundHeightDatum(
+          this.options.groundHeightDatum,
+          this.roots[id],
+          viewport as GeospatialViewport
+        )
+      });
       this._traverser.traverse(this.roots[id], frameState, this.options);
     }
   }
--- a/modules/tiles/src/tileset-3d/helpers/frame-state.ts
+++ b/modules/tiles/src/tileset-3d/helpers/frame-state.ts
@@ -18,6 +18,18 @@ export type FrameState = {
   sseDenominator: number; // Assumes fovy = 60 degrees
 };
 
+/** Options for {@link getFrameState}. */
+export type GetFrameStateOptions = {
+  /**
+   * Elevation in meters of the viewport's ground plane (z = 0) above the WGS84 ellipsoid.
+   * The viewport's `unprojectPosition` returns heights above the ground plane, while
+   * `region` bounding volumes carry absolute ellipsoidal heights, so the culling camera and
+   * frustum planes are lifted by this amount to align the culling frame with the content.
+   * Defaults to 0 (ground plane on the ellipsoid).
+   */
+  groundHeightDatum?: number;
+};
+
 const scratchVector = new Vector3();
 const scratchPosition = new Vector3();
 const cullingVolume = new CullingVolume([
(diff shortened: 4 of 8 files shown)


ibgreen:
Responded in issue. Will close for now, lets reopen if we resolve the cocerns.

PR #3637, opened 2026-08-21
feat(gltf): add Mesh Arrow primitive projection

ibgreen:
## Goals

- Add an opt-in, native glTF-to-Mesh-Arrow projection without replacing the existing glTF parser or changing normal loader behavior.
- Preserve scene placement and accessor semantics instead of flattening geometry implicitly.
- Consolidate the current Mesh Arrow contract and renderer-boundary documentation.

## Changes

- Add @loaders.gl/gltf/mesh-arrow with extractGLTFMeshArrowPrimitives().
- Return one Mesh Arrow table per primitive with source-compatible attribute descriptors, world matrix, node path, mesh/primitive indices, and material index.
- Preserve dense non-interleaved vertex buffer views; reject sparse and interleaved accessors explicitly rather than silently materializing them.
- Support point, triangle, and triangle-strip topology; document non-goals for transforms, skinning, morph targets, and GPU instancing.
- Add cross-runtime coverage for zero-copy backing storage, scene transforms, indexed primitives, and explicit unsupported-accessor errors.
- Add the Mesh Arrow usage guide and ArrowJS navigation, covering the table contract, luma.gl interop, GeoArrow boundary, instancing, packed primitives, and renderer-owned scheduling.

## Relationship to #3160

This is the native, narrowly scoped successor to #3160's GLBArrowLoader experiment. It retains the proposal's useful unit of output—one glTF primitive plus its world transform—while using loaders.gl's existing parser and I/O path, correctly naming the feature as glTF rather than the generic GLB container, and making normalization an explicit future concern instead of silently unwelding, de-instancing, or dequantizing geometry.

## Relationship to #3161

This incorporates #3161's Mesh Arrow documentation and its key architectural boundary: Mesh Arrow represents one renderable primitive, while instance data and draw scheduling remain separate Arrow- and renderer-level concerns. The guide also documents how the new glTF projection fits that boundary.

## Validation

- yarn lint fix
- yarn build
- yarn test-node
- yarn test-headless modules/gltf/test/lib/api/gltf-mesh-arrow.cross.spec.ts modules/obj/test/obj-loader.spec.js modules/pcd/test/pcd-loader.spec.ts
- yarn test-audit
- yarn test-website

The full headless suite currently also discovers a CJS ts-node fixture under test/apps/typescript-test/node_modules and fails before its tests run; the focused Chromium coverage passes.

### DIFF
--- a/docs/arrowjs/arrow-sidebar.json
+++ b/docs/arrowjs/arrow-sidebar.json
@@ -22,6 +22,11 @@
     },
     {
       "type": "category",
+      "label": "Usage Guide",
+      "items": ["arrowjs/usage-guide/mesharrow"]
+    },
+    {
+      "type": "category",
       "label": "Developer Guide",
       "items": [
         "arrowjs/developer-guide/introduction",
--- a/docs/arrowjs/usage-guide/mesharrow.md
+++ b/docs/arrowjs/usage-guide/mesharrow.md
@@ -1,0 +1,150 @@
+# Mesh Arrow Tables
+
+Mesh Arrow is the loaders.gl convention for representing one renderable mesh or point-cloud primitive as an Apache Arrow table. It is a vis.gl interoperability contract, not an Apache Arrow, GeoArrow, or glTF standard.
+
+The contract deliberately covers geometry attributes, primitive topology, indices, and a small amount of metadata. Scene graphs, materials, textures, instances, and draw scheduling remain separate concerns.
+
+## Data model
+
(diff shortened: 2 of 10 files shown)


ibgreen:
Superseded by #3695, which prototypes the revised convertGLTFToMeshArrow() API. Remaining glTF projection, extension coverage, batching, and Mesh Arrow contract work is tracked in #3700.

PR #3660, opened 2026-08-26
feat(las): decode typed Extra Bytes columns

ibgreen:
## Goals

- Expose descriptor-defined Extra Bytes as typed Arrow attributes.
- Preserve the existing raw `EXTRA_BYTES` output as the default.
- Keep complete LAS, complete LAZ, and chunked `parseInBatches` behavior aligned.
- Correctly represent LAS Extra Bytes scalar and vector descriptor metadata.

## Changes

- Add `las.extraBytes: 'raw' | 'typed'`, defaulting to `raw`.
- In typed mode, emit deterministic `EXTRA_BYTES_<descriptor-name>` attributes for supported numeric descriptors.
- Support scalar data types 1-6 and 9-10, plus deprecated 2-component and 3-component vector codes based on those scalar types.
- Parse all three scale and offset doubles from each 192-byte Extra Bytes descriptor and apply them per component.
- Use Float64 output for scaled or offset values so fractional and out-of-range transformed values are preserved.
- Select Extra Bytes by default when `las.extraBytes: 'typed'` is set without a `columns` list; raw Extra Bytes remain explicitly selectable through `las.columns`.
- Keep 64-bit integer descriptors and vectors based on them raw-only with a clear error in typed mode.
- Decode typed values from uncompressed records and the raw parity path for LAZ chunks, including split streaming batches.
- Share one complete raw LAZ chunk decode across waveform references, raw Extra Bytes, and typed Extra Bytes projections.
- Document naming, descriptor semantics, supported types, selection behavior, and limitations.

## Validation

- `yarn test-headless modules/las/test/las-loader.spec.ts`: 17 passed.
- `yarn test-slow modules/las/test/las-loader.node.slow.spec.ts`: 67 passed, 1 skipped.
- LAS TypeScript project check passed with `tspc --noEmit -p modules/las/tsconfig.json`.
- `yarn lint fix` passed.
- `git diff --check` passed.
- Repository build previously reached and built LAS successfully, then stopped in unrelated `modules/traces` because `@bufbuild/protobuf` was not installed in the dependency set.


ibgreen:
Addressed the review findings in eff5ce1b0. Typed Extra Bytes now parse descriptor scale/offset fields and apply them in both LAS and LAZ paths; the regression covers both formats. Added TSDoc to all new typed Extra Bytes helpers. The two review threads have been resolved.

ibgreen:
@codex review

ibgreen:
Addressed the fresh review findings in bb3da0c46. Scaled and offset integer (and scaled float32) Extra Bytes now use Float64Array output buffers, preserving fractional and out-of-range transformed values; regression coverage includes 250 + 7 and a fractional vector component. When `las.extraBytes: "typed"` is set without `las.columns`, Extra Bytes are now selected by default and covered by a dedicated test. Both review threads are resolved.

ibgreen:
@codex review

ibgreen:
Addressed both latest review comments in `86dead4e3`.

- The direct Arrow LAZ path now decodes a complete raw chunk once when waveform references, raw Extra Bytes, or typed Extra Bytes are requested, and shares that buffer across all projections. Added PDRF 9/10 complete and chunked parity coverage for waveform plus typed Extra Bytes.
- Documented the intentional selection rule: raw `EXTRA_BYTES` requires listing `EXTRA_BYTES` in `las.columns`; typed Extra Bytes are included when `las.extraBytes: 'typed'` is used without an explicit column list.

Focused LAS tests pass: 17/17, plus TypeScript checking, formatting, and diff checks.

PR #3663, opened 2026-08-26
feat(parquet): integrate Bloom-filter row-group pruning

ibgreen:
## Goals

Continue the Parquet selective-scan tranche after #3654 landed by making Bloom-filter metadata actionable for remote row-group reads.

## Changes

- Add canonical Parquet PLAIN encoding for Bloom-filter probe values.
- Decode and validate split-block Bloom-filter headers.
- Fetch Bloom-filter ranges during ParquetSource row-group planning.
- Prune only proven-negative equality/IN candidates.
- Preserve exact predicate evaluation after decoding.
- Add Bloom-filter read/byte/pruning telemetry.
- Add filter-column-first late materialization for the caller-thread path: predicate columns are decoded first, and projected columns are decoded only for matching rows.
- Keep worker paths conservative and unchanged until transferable row-index handling is integrated.
- Skip unsupported values, headers, disjunctions, malformed filters, and encrypted payloads conservatively.

## Remaining tranches

- Worker-side late materialization and shared range scheduling.
- High-cardinality scale benchmarks.
- Additional compatibility fixtures, including encrypted and malformed payloads.

## Validation

- yarn lint fix
- Focused Chromium ParquetSource tests: 15 passed.
- Full yarn build reaches the existing repository-wide missing @bufbuild/protobuf dependency errors in the traces package; no Parquet TypeScript errors were reported.


ibgreen:
Addressed all three Bloom-filter correctness findings in a992a90f8: masked the XXH64 lane product to 64 bits, hash BYTE_ARRAY values as raw payload bytes, and convert DECIMAL logical values to their physical representation before Bloom encoding. Added regression vectors; focused Parquet tests pass (22/22). The review arrived as one issue comment rather than resolvable inline threads.

PR #3669, opened 2026-08-26
docs(scan): expand format capability roadmap

ibgreen:
## Goals

Expand the common scan roadmap beyond Parquet and Iceberg to cover spatial, vector, transactional, streaming, and raster-adjacent formats.

## Changes

- Add explicit tranches for FlatGeobuf, GeoArrow/Arrow IPC, ORC, CSV/JSONL, Delta, Lance, GPU, and raster scans.
- Add a capability matrix distinguishing physical units and pushdown strength.
- Clarify that adapters report residual filtering when decoding is required.
- Keep TableQuery narrow and define the sibling RasterQuery direction.

This is documentation-only.

### DIFF
--- a/docs/developer-guide/common-scan-architecture.md
+++ b/docs/developer-guide/common-scan-architecture.md
@@ -481,25 +481,64 @@ semantics?
 
 ## Roadmap
 
-The implementation evolves in reviewable tranches. The foundation now provides the first four
-items; the later work can land independently without changing portable query semantics:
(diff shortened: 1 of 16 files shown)


ibgreen:
Addressed the inline review feedback in 8312f3593. Added the explicit pushdown+residual capability level and updated Parquet/Iceberg to advertise it, so explain output preserves metadata/page pruning while documenting exact residual predicate evaluation. Updated the roadmap and capability tests accordingly.

PR #3667, opened 2026-08-26
feat(3d-tiles): support multiple tile contents

ibgreen:
## Goals

- Extend loaders.gl 3D Tiles compatibility to tiles that declare multiple content payloads.
- Publish a clear format compatibility matrix for parser, traversal, extension, metadata, and renderer-facing support.

## Changes

- Accept and validate 3D Tiles 1.1 content arrays.
- Preserve content metadata and resolved URLs in source order.
- Load multiple explicit content payloads together and expose them through `Tile3D.contents`.
- Preserve the legacy primary `Tile3D.content` field for existing consumers.
- Account for all loaded payloads in memory usage, nested-tileset installation, compression flags, and point statistics.
- Add focused Chromium tests for normalization and runtime loading.
- Add the 3D Tiles format compatibility page, sidebar entry, and `whats-new` entry.

## Validation

- `yarn lint fix`
- Focused Chromium test: `modules/3d-tiles/test/multiple-contents.browser.spec.ts`

The full repository build is currently blocked by pre-existing missing `@bufbuild/protobuf` dependencies in `modules/traces`.

Related tracker: #1245

PR #3671, opened 2026-08-26
feat(3d-tiles): preserve implicit subtree metadata

ibgreen:
## Goals

- Preserve implicit-subtree metadata references during lazy materialization.
- Make the current metadata capability explicit without claiming value decoding that is not implemented.

## Changes

- Carry subtree property tables, tile metadata, content metadata, and subtree metadata through the parsed implicit-subtree model.
- Attach immutable-by-convention metadata references to generated implicit tile headers as `implicitMetadata`.
- Preserve metadata on both materialized tiles and lazy subtree-boundary placeholders.
- Add focused Chromium coverage and document the runtime limitation: metadata values and classes are not decoded yet.
- Update the implicit-tiling guide and `whats-new`.

## Validation

- `yarn lint fix`
- Focused Chromium tests: implicit metadata plus existing implicit traversal tests (8/8)

Related tracker: #1245

### DIFF
--- a/docs/modules/3d-tiles/concepts/implicit-tiling-and-subtrees.md
+++ b/docs/modules/3d-tiles/concepts/implicit-tiling-and-subtrees.md
@@ -160,7 +160,9 @@ The underscored SSE field is diagnostic rather than stable API. Use these values
 ## Current Limitations
 
 - Only the first `contentAvailability` stream is used; `3DTILES_multiple_contents` is not implemented.
-- Subtree and tile metadata semantics beyond availability are not materialized yet.
+- Subtree metadata references (`propertyTables`, `tileMetadata`, `contentMetadata`, and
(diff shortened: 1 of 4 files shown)


ibgreen:
Rebased onto current master (d15c62a04) and force-pushed. The prior workflow had all test jobs passing; its build job was reported before the workflow completed, so this push starts a fresh validation run. Focused implicit metadata and traversal tests pass 8/8. I also checked merged PR #3667: it has no open inline review threads or unresolved review requests.

ibgreen:
CI fix and rebase: the build failure was TS2345 because normalizeImplicitTileData receives the 3D Tiles Subtree type, whose metadata fields are intentionally unknown-shaped. ParsedImplicitSubtree metadata references now use unknown-compatible types while retaining the same runtime payload. Rebased onto current master and force-pushed caff44686. Focused tests remain 8/8.

PR #3678, opened 2026-08-26
feat(copc): stream node ranges and hierarchy pages

ibgreen:
## Goals

- Advance COPC streaming beyond full-node range fetches for the TypeScript position-only path.
- Expose hierarchy pages as an async stream while preserving the existing hierarchy cache and tile APIs.
- Keep RGB decoding and atomic compatibility behavior unchanged until progressive RGB readiness is implemented.

## Changes

- Fetch TypeScript position-only COPC node data through sequential byte ranges controlled by `rangeChunkSize`.
- Feed those ranges into the LAZ decoder and emit positions as soon as the PDRF 6-8 Point14 layer is available.
- Preserve complete-range decoding for requests that include RGB.
- Add `COPCTileSource.loadHierarchyInBatches(options)` with cancellation and `maxPages` support.
- Merge streamed hierarchy pages into the existing source cache.
- Add public `COPCHierarchyBatch` and `COPCHierarchyBatchOptions` types.
- Extend the feedable LAZ decoder with progressive position-batch state and remaining-point tracking.
- Add focused COPC tests for chunked position ranges and hierarchy page streaming.
- Update COPC API documentation with current progressive limitations.

## Validation

- `yarn test-headless modules/copc/test/copc-source.spec.ts`: 14 passed.
- `tspc -b modules/loader-utils/tsconfig.json modules/las/tsconfig.json modules/copc/tsconfig.json --pretty false`: passed.
- `yarn lint fix` passed.
- `git diff --check` passed.

## Limitations

- Progressive RGB output still requires a complete node range.
- LASzip variable-chunk table streaming and arbitrary sparse layer fetches remain follow-up work.

### DIFF
--- a/docs/modules/copc/api-reference/copc-source-loader.md
+++ b/docs/modules/copc/api-reference/copc-source-loader.md
@@ -35,11 +35,13 @@ The created data source exposes the point-cloud tile methods used by `PointCloud
 - `getChildren(tile)` returns available child tile headers.
 - `loadTileContent(tile)` returns normalized point positions, optional colors, point count, and cartographic origin.
 - `loadTileContentInBatches(tile, options)` yields normalized Arrow point tables progressively after the selected node range has been fetched. The TypeScript LAZ variant is required. `options.batchSize` controls the maximum points per table, `options.columns` can request positions only, and `options.signal` cancels the request/decode.
+- `loadHierarchyInBatches(options)` yields hierarchy pages and their discovered nodes as they are fetched.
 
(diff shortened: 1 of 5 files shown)


ibgreen:
Addressed all three review comments in `8bab625bd`.

- Rewrote the newly added hierarchy-page test with native Vitest `test`/`expect`.
- Removed consumed child-page placeholders from `_hierarchy.pages`, preventing duplicate hierarchy range requests through later `getChildren()` calls.
- Rechecked cancellation after every awaited node-range request before yielding the returned bytes.

Validation: focused COPC suite 14/14, project-reference type check passed, lint and diff checks passed.

PR #3881, opened 2026-08-28
feat(draco): add per-attribute quantization

ibgreen:
## Goals

- Let applications tune quantization independently for attributes that share a Draco category.
- Expose the official Draco ExpertEncoder path without changing existing category-level writer behavior.
- Reject malformed quantization settings before native encoding.

## Changes

- Adds typed `draco.attributeQuantization` options keyed by application attribute name.
- Supports either a bit count or an explicit `{bits, origin, range}` quantization transform.
- Expands existing category defaults by Draco attribute ID and lets exact attribute settings override them.
- Corrects and exports the vendored `ExpertEncoder` TypeScript declaration.
- Validates bit depth, attribute names, explicit origin dimensions and values, and explicit ranges.
- Adds deterministic builder coverage plus a real browser/WASM round trip proving independent `TEXCOORD_0` and `TEXCOORD_1` transforms.
- Documents the new option and exports its public types.

## Validation

- `yarn install`
- `yarn lint fix`
- `yarn build`
- `yarn build-workers`
- `yarn test-node` (361 passed, 6 skipped)
- `yarn test-headless` (3258 passed, 39 skipped)
- focused Draco headless tests (28 passed)
- `yarn test-audit` (0 violations)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### v5.0.0-alpha.2
 
+- feat(draco): add validated per-attribute quantization and explicit quantization transforms to `DracoWriter`
 - feat(gltf): select AVIF texture sources when the active image decoder supports them, and support independent `KHR_texture_transform` transforms for textures that share UV data (#3611, #3632)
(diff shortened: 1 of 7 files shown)


ibgreen:
Addressed in stacked PR #3884: NORMAL quantization is restricted to native-safe 2–29 bits; explicit NORMAL transforms and ineffective integer quantization are rejected.

PR #3884, opened 2026-08-28
feat(draco): optimize glTF decode and expose GPU metadata

ibgreen:
## Goals

Extend the Draco/glTF path with a smaller glTF-optimized decoder, preserve quantization metadata for GPU consumers, support selective attribute extraction, and provide encoding diagnostics.

## Changes

- Add the official Draco 1.5.7 glTF decoder assets and `decoderProfile: 'gltf'` (the KHR Draco glTF extension selects it by default; `'full'` remains available).
- Preserve normalized and quantization/octahedral transforms through Mesh Arrow schema metadata and conversion.
- Add include/exclude attribute selection with validation and transform-aware bounds.
- Add `encodeSyncWithReport()` and public `encodeDraco()` diagnostics.
- Harden quantization validation for integer and NORMAL attributes.

The branch includes the predecessor quantization work from #3881; this PR supersedes that branch and should be merged after its review is complete.


ibgreen:
Addressed: decoder profile initialization is now serialized to prevent wrapper/WASM cross-talk under concurrent loads. Arrow-to-mesh conversion restores serialized logical bounds when available.

PR #3885, opened 2026-08-28
feat(draco): add batch encoding API

ibgreen:
## Goals

Add a throughput-oriented batch Draco writer path for applications encoding many geometries without repeatedly initializing the runtime.

## Changes

- Add public `encodeDracoBatch()` that loads and initializes Draco once.
- Process geometries sequentially to bound peak native memory and avoid retaining native geometry objects between results.
- Return the same per-geometry `DracoEncodingResult` diagnostics as `encodeDraco()`.
- Add a focused writer test, benchmark coverage, and API documentation.

This follow-up is stacked on #3884.


ibgreen:
Addressed: batch benchmark throughput now counts both geometries and applies each labeled Draco option.

PR #3888, opened 2026-08-28
feat(i3s): support vertical CRS and elevation placement

ibgreen:
## Goals

- Complete I3S vertical CRS and elevation semantics for mesh and Point Cloud sources.
- Align the implementation with the canonical math.gl CRS architecture landed on `master`.
- Make terrain/scene placement, units, height datums, bounds, and initialization behavior explicit and documented.

## Changes

- Normalize I3S `heightUnit`, legacy `ZFactor`, and `elevationInfo.unit` values with common ArcGIS unit aliases.
- Apply `absoluteHeight`, `onTheGround`, `relativeToGround`, and `relativeToScene` formulas to geometry, origins, traversal bounds, and target-space bounds.
- Add batched `TilesetElevationProvider` contracts for WGS84 terrain and scene sampling, including provider unit and height-reference conversion through registered math.gl geoid models.
- Require provider-wide conservative elevation ranges for traversal bounds so interior peaks and depressions cannot be culled.
- Preserve Float64 precision through placement and transformation, and keep metric bound extents from being scaled with source vertical units.
- Add async tileset preparation so root bounds are placed before runtime tile initialization.
- Use `@math.gl/crs` `SpatialReference` as the canonical immutable source for CRS, vertical CRS, coordinate order, frame, epoch, and per-component units.
- Route horizontal/geocentric transformations through `@math.gl/proj4` and height-datum conversions through `@math.gl/geoid`; retain only I3S-specific unit, placement, provider, and geometry semantics in the I3S adapter.
- Reclassify caller CRS overrides, transform projected normals with an inverse-transpose local Jacobian, and reject every unresolved format-owned spatial operation.
- Add focused mesh, Point Cloud, unit, provider, geoid, bounds, metadata-preservation, and failure tests.
- Add the framework vertical CRS developer guide and update I3S/3D CRS documentation, feature matrix, roadmap, and release notes.
- Merge the latest `master`, including the math.gl alpha.6 spatial-reference foundation.

## Validation

- `yarn lint fix`
- `yarn build-modules`
- `yarn test-node` (372 passed, 6 skipped)
- `yarn test-headless` (3381 passed, 39 skipped)
- focused CRS/I3S Chromium suite (63 passed)
- `yarn test-audit` (0 violations)

## Notes

- Surface providers are application supplied; loaders.gl never downloads terrain implicitly.
- Per-feature Arcade/renderer expression evaluation remains in the renderer and popup tranche.
- Compatibility aliases remain available, but canonical CRS discovery metadata is owned by the math.gl `SpatialReference` descriptor.


ibgreen:
Addressed all four findings in 88a35ae7a7:

- Unresolved I3S spatial descriptors now reject unconditionally in both the mesh and Point Cloud source paths, including invalid Z factors, unsupported vertical units, and unsupported elevation modes.
- The elevation-provider contract now includes `getElevationRange()`. Bounds use its conservative extrema over the complete dateline-aware footprint, rather than assuming corner point samples cover interior peaks or depressions. Invalid ranges and incomplete providers reject explicitly.
- A caller `sourceCrs` override now reclassifies the canonical coordinate frame from the effective CRS and drops stale per-component units from the original CRS.
- Projected normals now use the inverse transpose of a centered finite-difference ECEF-to-target Jacobian, with a singularity check.

Regression coverage exercises both source rejection paths, interior surface extrema, invalid/incomplete range providers, source-frame override reclassification, and mixed horizontal/vertical projected normals.

Validation completed locally:

- `yarn lint fix`
- `yarn build-modules`
- `yarn test-node` — 372 passed, 6 skipped
- `yarn test-headless` — 3381 passed, 39 skipped
- focused CRS/I3S Chromium suite — 63 passed
- `yarn test-audit` — 0 violations

PR #3950, opened 2026-09-01
feat(i3s): decode LEPCC point clouds in workers

ibgreen:
Goals

- Move I3S Point Cloud LEPCC geometry and attribute decoding off the renderer thread when workers are available.
- Allow Potree LAS/LAZ content to use the worker-capable TypeScript LAS loader by default.

Changes

- Add the standalone I3SLEPCCLoader metadata/parser pair with browser and Node worker bundles.
- Expose LEPCC loader options and package worker entry points.
- Route I3SPointCloudSource LEPCC resources through the injected CoreAPI, with a decoder fallback when workers are unavailable or disabled.
- Pass CoreAPI through I3S Point Cloud source creation.
- Preserve worker and fetch settings in Potree LAS/LAZ node loads instead of forcing core.worker: false.
- Keep legacy laz-perf, COPC, and laz-rs compatibility variants main-thread-only.
- Add worker/source coverage and update I3S format documentation.

Validation

- yarn build
- yarn build-workers
- yarn lint fix
- yarn test-audit
- yarn test-node
- yarn test-headless
- Targeted I3S LEPCC, I3S Point Cloud, Potree, and LAS worker tests

### DIFF
--- a/docs/modules/i3s/formats/i3s.md
+++ b/docs/modules/i3s/formats/i3s.md
@@ -174,7 +174,7 @@ those versions.
 
 | Capability | Status | Since | Notes |
 | --- | :---: | --- | --- |
-| LEPCC point-cloud attribute blobs | **Supported** | **v5.0** | `I3SLEPCCDecoder` decodes standalone `lepcc-xyz`, `lepcc-rgb`, `lepcc-intensity`, and bit-stuffed or Huffman flag-byte resources. `I3SPointCloudSource` maps them to Arrow point attributes. |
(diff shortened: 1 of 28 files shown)


ibgreen:
Addressed in commit 730d45b35f:\n\n- Added browser and Node LEPCC worker build scripts and included them in the I3S build.\n- Split the metadata loader into i3s-lepcc-loader-types.ts.\n- Made the public i3s-lepcc-loader subpath parser-bearing, with the metadata loader preloading it through the static package subpath.\n- Added the TypeScript path mapping needed for the package subpath.\n\nValidation passed: yarn build, yarn lint fix, I3S focused tests (15), Potree tests (8), and LAS/LAZ worker tests (12).\n\nThe GeoArrow vertex-count note is outside this PR's changed files and should be tracked separately.

ibgreen:
Fixed the failing Node 24 headless check in 6489e26fd8. Potree's LAS/LAZ path now omits an undefined `core` options block so the worker-capable TypeScript LAS loader can use its packaged worker by default, while explicit source `core` settings are preserved. Updated the boundary test to cover both defaults and an explicit `worker: false` override. Focused browser test, lint, and build pass locally.

PR #3979, opened 2026-09-09
feat(stac): support Portolan catalogs

ibgreen:
## Goals

Support Portolan as a STAC publishing profile and demonstrate the metadata path in the browser-based Overture example.

## Changes

- Add Portolan profile detection and version extraction to `@loaders.gl/stac`.
- Expose Portolan metadata from `STACSource.getMetadata()`.
- Support collection-level assets for Portolan single-file collections.
- Update STAC URL detection and documentation.
- Inspect the Overture STAC hierarchy for a Portolan declaration in the Overture example.
- Load same-origin absolute worker URLs directly so Vite localhost workers do not require a Blob `importScripts()` wrapper.

## Validation

- `yarn lint fix`
- `yarn test-audit`
- `yarn test-headless modules/stac/test/stac-source.spec.ts modules/worker-utils/test/lib/worker-utils/get-loadable-worker-url.spec.ts`
- `yarn build`
- `cd examples/website/overture-parquet && yarn build`

The current authoritative Overture STAC catalog does not declare Portolan, so the example only displays profile metadata when Overture publishes it.

### DIFF
--- a/docs/modules/stac/api-reference/stac-source-loader.md
+++ b/docs/modules/stac/api-reference/stac-source-loader.md
@@ -71,6 +71,9 @@ Returns the root Catalog or Collection, declared conformance classes, and the de
 
 The root document is fetched once and cached.
 
+When the root declares the Portolan publishing profile, the returned metadata also includes
+`portolanExtension` and `portolanVersion`.
+
 ### `getCollections(options?)`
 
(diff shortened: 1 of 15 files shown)


ibgreen:
@jatorre

